PR #32, opened 2026-06-12
chore: disable Jekyll processing for GitHub Pages

rijul-mck:
Pages is now enabled (serving main /). The .nojekyll marker makes it serve the tree statically so systems/superintelligence/HOW-IT-WORKS.html — the SI orchestrator guide — renders as-is at https://rkz91.github.io/coco/systems/superintelligence/HOW-IT-WORKS.html.


coco-research (approved):
INDEX regenerated, checks green — approving.

PR #36, opened 2026-06-30
Add Cognee knowledge graph memory bundle

coco-research:
# Add Cognee Knowledge Graph Memory Bundle

Adds Cognee as an optional memory backend to Coco, following the existing system bundle pattern (like Brain, GSD, Team).

### New: `systems/cognee/` bundle (3 skills)

| Skill | Purpose |
|-------|---------|
| `/cognee` | Status check, dataset init, backend switching (brain ↔ cognee) |
| `/cognee-store` | Push entities, decisions, events, relationships into the knowledge graph. End-of-session flush. |
| `/cognee-recall` | Semantic + graph search across projects. Context injection for informed agent decisions. |

## Why Cognee

[Cognee](https://github.com/coco-research/cognee) is an open-source AI memory platform that gives agents persistent long-term memory via a self-hosted knowledge graph engine. Compared to the existing Brain bundle (SQLite):

| Feature | Brain (SQLite) | Cognee |
|---------|---------------|--------|
| Storage | Flat relational DB | Knowledge graph + embeddings |
| Search | Keyword + relational | Semantic + graph traversal + lexical (14 modes) |
| Cross-project | Manual aggregation | Native cross-dataset queries |
| Sessions | Stateless | Multi-turn session tracking |
| Auto-recall | Manual context query | Automatic memory injection |

## Design

- **Coexists with Brain** — not a replacement. Users pick one or both.
- **Zero new dependencies** — skills are pure Markdown with `curl` commands to Cognee's HTTP API.
- **Zero core changes** — follows the exact same pattern as `systems/brain/`. The install script auto-discovers any `systems/<name>/skills/` directory.
- **Graceful fallback** — if Cognee isn't running, skills offer to use Brain instead.

## Install

```bash
pip install cognee && cognee server start   # start Cognee
bash install.sh --adapter claude-code --systems cognee   # wire into Coco
```

## File count

- 3 SKILL.md files (cognee, cognee-store, cognee-recall)
- 1 README.md (bundle overview)
- README.md edits (system bundle count: 4→5, skills count: 146→149, added Cognee deep-dive section, updated demo)
- Zero changes to core Coco code or install scripts

PR #31, opened 2026-06-10
docs(readme): rewrite header to match catchy Apple-style tagline

coco-research:
Matches the catchy style applied to the About description.

PR #38, opened 2026-07-03
security: bootstrap confirmation gate, Cursor adapter safety, prompt-injection hardening

imachiever:
## Summary

Three fixes that touch runtime behavior — each deliberately scoped to preserve the documented contract wherever possible. Part 2 of 3 PRs split by risk tier (see #37 for the "merge today" safe-hardening tier); this one deserves a slower, more deliberate look since one item is a genuine, intentional behavior change.

- **Bootstrap confirmation gate** (`bin/coco-bootstrap.sh`) — the `bash <(curl -fsSL ...)` one-liner previously git-cloned and immediately ran `install.sh` with zero pause and zero visibility into what was about to execute. Now displays the cloned commit's hash/date/message and asks `[y/N]` before proceeding. **This is an intentional, documented behavior change** — pass `--yes` or set `COCO_BOOTSTRAP_YES=1` to preserve the old unattended flow for CI/scripted installs. Documented in `SECURITY.md` (deliberately not `README.md` — that file is independently contended by two other open PRs upstream, #31 and #36, and touching it here would create an avoidable merge conflict).
- **Cursor adapter safety** (`adapters/cursor/install.sh`) — brought `link_dir()`/`run()` in line with the Claude Code adapter's already-safer versions: guards against silently overwriting a real file (not just a symlink) at the destination, and replaces a `shellcheck SC2015`-flagged `A && B || C` pattern with explicit `if/then/else` (the old form could execute the real command even in dry-run mode if the `echo` failed, e.g. on `SIGPIPE`).
- **Prompt-injection hardening** (`skills/find-skills/SKILL.md`) — Step 4 documents `npx skills add <owner/repo@skill> -g -y`, where `-y` skips the CLI's own confirmation and the package identifier comes entirely from search results (skills.sh, GitHub) that an attacker can influence by publishing a well-named malicious package. **Deliberately did not remove `-y` from the documented command** — that would change what a human running it directly in a terminal experiences, a real change to a documented CLI invocation. Instead added agent-instruction guidance only: before the agent invokes the install on the user's behalf, it must show the resolved package identity and get explicit confirmation referencing that specific package. Zero change to the documented command, flag, or CLI output.

## Test plan

- [x] `bash -n` syntax check on all shell scripts
- [x] `bash tests/check-command-refs.sh` / `check-evidence-gate.sh` — PASS
- [x] INDEX drift check — zero drift (confirmed the find-skills edit doesn't touch `description:` frontmatter, only body content)
- [x] Functional test of the bootstrap gate's accept path (`COCO_BOOTSTRAP_YES=1`) — confirmation skipped, install proceeds to completion
- [x] Functional test of the bootstrap gate's decline path (piped `n`) — prompt shown, correctly aborts before `install.sh` runs, clean exit
- [x] Cursor adapter dry-run smoke test — output shape unchanged from before the fix
- [x] Independent re-verification from a fresh clone — matches all claims above

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -31,6 +31,17 @@ Coco is a library of markdown artifacts and shell installers. The most likely se
 | Hardcoded credentials | high | Coco ships zero secrets; report any you find immediately |
 | Supply-chain via plugin recommendations | low | We link external plugins but don't bundle them |
 
+## Bootstrap installer behavior
+
+`bin/coco-bootstrap.sh` (the `bash <(curl -fsSL ...)` one-liner) now pauses for interactive
+confirmation before installing — it displays the cloned commit's hash, date, and message, and
+asks `[y/N]` before running `install.sh`. This is a deliberate behavior change from the prior
+zero-friction flow, added so users piping a remote script directly to their shell get a chance
+to verify what they're about to run.
+
+Pass `--yes` or set `COCO_BOOTSTRAP_YES=1` to preserve the old unattended behavior for CI or
+scripted installs.
+
 ## Out of scope
 
 - Issues in third-party AI tools (Claude Code, Cursor, Codex, etc.) — report to those projects
--- a/adapters/cursor/install.sh
+++ b/adapters/cursor/install.sh
@@ -20,10 +20,13 @@ while [[ $# -gt 0 ]]; do
   shift
 done
 
-run() { [[ $DRY_RUN -eq 1 ]] && echo "DRY: $*" || "$@"; }
+run() {
+  if [[ $DRY_RUN -eq 1 ]]; then echo "DRY: $*"; else "$@"; fi
+}
 
 link_dir() {
   local src=$1 dst=$2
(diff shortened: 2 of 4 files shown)


coco-research:
Landed via #49, now merged. Thank you — the confirmation gate, both Cursor adapter shell fixes, and the prompt-injection guidance all went in unmodified with your authorship preserved.

I added one follow-up commit on top for two things worth calling out. First, the remaining `rkz91` URLs are now `coco-research`: that account renamed on 2026-07-18 and an unrelated third party registered the vacated username the day before, so the documented curl-pipe-to-bash install line and the "verify this commit" link would have pointed at content they control if they ever create a repo named `coco`. Accurate when you wrote it, worth fixing now. Second, `COCO_BOOTSTRAP_YES=0` skipped the prompt because the guard tested `-z` rather than the value; only `1` or `--yes` skips it now.

Closing in favour of #49. Sorry for the extra branch: the PR could not be updated in place from here.

PR #42, opened 2026-07-18
chore(release): cut v1.1.0

coco-research:
Bumps the release version 1.0.0 → 1.1.0 to match the Spec Version bump and the rebrand + open-core relicense.

- README version badge, `package.json`, `.claude-plugin.json`, `Formula/coco.rb` (version + tag URL) → **1.1.0**
- CHANGELOG `[Unreleased]` → **`[1.1.0]` — 2026-07-18**

**Follow-up in this same flow:** the Formula `sha256` is a placeholder here. After this merges I'll tag `v1.1.0`, create the GitHub release, compute the real tarball checksum, and open a second PR to set it (mirrors how #30 set the v1.0.0 sha).

PR #43, opened 2026-07-18
fix(formula): set real sha256 for v1.1.0 tarball

coco-research:
Sets the Homebrew formula `sha256` to the real checksum of the published `v1.1.0` source tarball (was a placeholder in #42).

- Tarball: https://github.com/coco-research/coco/archive/refs/tags/v1.1.0.tar.gz
- sha256: `e6024602266408f51f8fef2f7e4abc8d7c67c98ee5fae45b00ea799684d86cf0`

`brew install coco-research/coco/coco` now resolves to v1.1.0.


rijul-mck (approved):
Real sha256 for the published v1.1.0 tarball. CI green. Approving.

PR #48, opened 2026-07-31
security: SHA-pin Actions, least-privilege permissions, CSP, XSS fix, CVE floor bumps (rebase of #37)

coco-research:
This lands @imachiever's security work from #37. All four commits remain authored by **Rajat Bhatia <imachiever@gmail.com>**; this branch exists only because #37 could not be updated in place.

Supersedes #37. Please close that one in favour of this.

## Why it needed a rebase

#37 was opened on 2026-07-03 and had drifted into a real conflict. Merging it as drafted would have **regressed** `actions/setup-node` from **v7 back to v4** across three workflows, undoing the dependabot bump that already landed in #39.

## What I changed during conflict resolution

Only the Action pins. His SHA-pinning intent is preserved; the pins now target the versions `main` actually uses:

| Action | #37 as drafted | This branch |
|---|---|---|
| `setup-node` | `49933ea…` (**v4**, a regression) | `820762786026740c76f36085b0efc47a31fe5020` (**v7**) |
| `setup-python` | `a26af69…` (v5) | unchanged, `a26af69…` (v5) |

Both SHAs were verified two ways: each exists in the genuine upstream repository, and each matches what that tag currently points to. `actions/checkout` is deliberately left unpinned, as in the original PR.

## His changes, unmodified

- `permissions: contents: read` added to three workflows. Pure privilege narrowing.
- **Real DOM-XSS fix** in `skills/brainstorming/scripts/helper.js`: `innerHTML` with concatenated user-controlled label becomes `textContent`, with a null guard.
- CSP meta tag added to the generated diagram artifacts. Worth noting for reviewers that it retains `'unsafe-inline'` on `script-src` because the templates use inline modules, so it is a partial mitigation rather than a hard boundary.
- CVE floor bumps in the `openai-apps-mcp` template: `@modelcontextprotocol/sdk`, `hono`, `vite`, `wrangler`, all same-major.

## Verification

Every workflow parses as valid YAML, no conflict markers remain, `helper.js` passes `node --check`, the template `package.json` parses, and `check-command-refs.sh`, `check-evidence-gate.sh`, `install.sh` syntax and the index-freshness check all pass locally.

**Heads-up:** dependabot #44 proposes `setup-python` 5 → 7. Once these pins land, that PR will need its pin updated rather than just its version string.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [main]
 
+permissions:
+  contents: read
+
 jobs:
   lint-and-test:
     runs-on: ubuntu-latest
@@ -13,17 +16,17 @@ jobs:
       - uses: actions/checkout@v7
 
       - name: Set up Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
         with:
           python-version: '3.11'
 
       - name: Set up Node
-        uses: actions/setup-node@v7
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7
         with:
           node-version: '20'
 
       - name: Install pyyaml
-        run: pip install pyyaml
+        run: pip install pyyaml==6.0.2
 
       - name: Validate adapter manifests
         run: |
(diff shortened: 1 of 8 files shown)


rijul-mck (approved):
Approved. This preserves @imachiever's security work with his authorship intact and fixes the one blocking problem: as drafted, #37 would have reverted actions/setup-node from v7 to v4 across three workflows, undoing the bump from #39. Pins now target v7 and v5, and I verified each SHA both exists upstream and matches what that tag currently points to. CI passing here is meaningful because this PR edits the workflows themselves, so the pinned actions genuinely resolved and ran. The XSS fix and permission narrowing are unmodified.

PR #49, opened 2026-07-31
security: bootstrap confirmation gate + Cursor adapter safety (rebase of #38)

coco-research:
This lands @imachiever's work from #38. His three commits remain authored by **Rajat Bhatia <imachiever@gmail.com>**; this branch exists only because #38 could not be updated in place from here.

Supersedes #38. Please close that one in favour of this.

## His changes, unmodified

- **A confirmation gate in `bin/coco-bootstrap.sh`.** The installer is a curl-pipe-to-bash entry point, so it now clones first, prints the commit hash, date and message, and pauses for `[y/N]` before running `install.sh`. The default is fail-closed: anything other than `y`/`yes` aborts and tells you where the clone is so you can inspect it.
- **Two real shell bugs fixed in `adapters/cursor/install.sh`**: a `link_dir()` that could silently clobber a real file, and an `A && B || C` dry-run pattern that could execute the real command if the `echo` failed. Both are textbook SC2015 fixes.
- **Prompt-injection guidance** in `find-skills`, telling the agent to surface and confirm the resolved package name before running `npx skills add … -y` on someone's behalf.

## Two follow-ups I added, as a separate commit

**1. The remaining `rkz91` URLs are now `coco-research`.** This is the substantive one. That account renamed to `coco-research` on 2026-07-18 and an unrelated third party registered the vacated username the day before. Two of the three occurrences were the documented curl-pipe-to-bash install command and the "verify this commit" link, so if that account ever creates a repository named `coco`, both would have pointed at content they control. The third, `REPO_URL`, was already corrected on `main` and the rebase picked it up.

**2. `COCO_BOOTSTRAP_YES=0` used to SKIP the confirmation prompt.** The guard tested `-z "$YES"` (empty) rather than the value, so anyone setting `0` to mean "off" got the opposite of what they asked for. Now only the exact value `1`, or `--yes`, skips it.

Verified behaviourally: unset → prompt shown, `0` → prompt shown, `1` → prompt skipped.

## Verification

`bash -n` passes on the bootstrap script and every adapter installer. No `rkz91` references remain. `check-command-refs.sh`, `check-evidence-gate.sh` and the index-freshness check all pass locally.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -31,6 +31,17 @@ Coco is a library of markdown artifacts and shell installers. The most likely se
 | Hardcoded credentials | high | Coco ships zero secrets; report any you find immediately |
 | Supply-chain via plugin recommendations | low | We link external plugins but don't bundle them |
 
+## Bootstrap installer behavior
+
+`bin/coco-bootstrap.sh` (the `bash <(curl -fsSL ...)` one-liner) now pauses for interactive
+confirmation before installing — it displays the cloned commit's hash, date, and message, and
+asks `[y/N]` before running `install.sh`. This is a deliberate behavior change from the prior
+zero-friction flow, added so users piping a remote script directly to their shell get a chance
+to verify what they're about to run.
+
+Pass `--yes` or set `COCO_BOOTSTRAP_YES=1` to preserve the old unattended behavior for CI or
+scripted installs.
+
 ## Out of scope
 
 - Issues in third-party AI tools (Claude Code, Cursor, Codex, etc.) — report to those projects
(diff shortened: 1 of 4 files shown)


rijul-mck (approved):
Approved. The confirmation gate is fail-closed and the two Cursor adapter shell fixes are correct. The added commit matters for a real reason: it removes the last rkz91 URLs from a curl-pipe-to-bash installer, where that username is now held by an unrelated third party, and it fixes COCO_BOOTSTRAP_YES=0 silently skipping the very prompt it was meant to keep. Behaviour verified: unset and 0 both prompt, only 1 skips.

PR #46, opened 2026-07-31
skills: add scroll-world (scroll-scrubbed 3D world landing pages)

coco-research:
Adds [scroll-world](https://github.com/oso95/scroll-world) — build an immersive scroll-scrubbed landing page where a pre-rendered camera flies from outside each scene into its interior, then flows on to the next with no cuts: one continuous connected flight.

## Attribution

MIT, **Copyright (c) 2026 cyw** — upstream `LICENSE` retained at `skills/scroll-world/LICENSE`, matching this repo's per-skill licensing convention (as with `skills/skill-creator`, `skills/openai-agents`).

Upstream: `oso95/scroll-world` v0.8.0.

## Verified working, not just installed

The clip generation leg needs paid third-party credits (Monid / Seedance, Higgsfield fallback), so that was not exercised. What ships to the end user is the **portable framework-agnostic scroll-scrub engine**, and that was tested for real:

- Generated H.264 clips to the skill's own stated spec — `crf 20`, `-g 8`, `+faststart`, no audio
- Mounted `mountScrollWorld()` in a browser at 1280×860 with 3 scenes + 2 connectors
- Engine built a **7.00-viewport scroll track** (6020px), matching the configured dive/connector weights
- **Lazy-loaded all 5 clips as `blob:` URLs**, `readyState: 4` — as documented, so it does not depend on HTTP byte-range support
- **Scroll genuinely scrubbed `currentTime`** across clips in sequence and handed off between them: at 55% `[1.725, 1.725, 1.061, 0, 0]` → at 70% `[1.992, 1.992, 1.992, 0.881, 0]` → at 100% all `1.992`
- Nav active-state tracked scroll position, per-section `scroll`/`linger` pacing overrides applied, `atmosphere` particle layer rendered
- `scrub-engine.js` passes `node --check`; `knockout.py` passes `py_compile`; zero hardcoded outbound endpoints in the skill

## Gates

Leak scan · frontmatter lint · `check-command-refs.sh` · `check-evidence-gate.sh` · `install.sh` syntax · claude-code adapter dry-run (discovers the skill) · README untouched — all green locally.

`skills/INDEX.md` is regenerated because CI enforces it. `README.md` counts are deliberately left to a single reconciliation PR at the end of the series.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #45, opened 2026-07-30
skills: add coco-loop (governed autonomous loop)

coco-research:
Adds the `/coco-loop` skill: it turns a plain-language goal into a readable charter the person confirms, then arms and runs a bounded loop **propose-only**. It never commits on its own.

This is the first of a series of one-skill-per-PR additions, so each one stays independently reviewable and testable.

## What's in this PR

| File | Purpose |
|---|---|
| `skills/coco-loop/SKILL.md` | The skill |
| `skills/coco-loop/LICENSE` | Coco-Research Proprietary, retained per the repo's open-core model (as with `systems/superintelligence/LICENSE`) |

## One content change from the source

The Prerequisites section previously said to install "from `pip install -e` in the coco-loops repo" without giving a URL, which was a dead end for anyone outside the project. It now links [coco-research/coco-loops](https://github.com/coco-research/coco-loops) with clone and install steps.

## Verification

All gates were run locally before opening this PR:

- **Leak scan** — no absolute home paths, internal hosts, or secrets
- **Frontmatter lint** (same logic as `lint-frontmatter.yml`) — 139 `SKILL.md` files valid, including this one
- **`tests/check-command-refs.sh`** — pass
- **`tests/check-evidence-gate.sh`** — pass
- **`install.sh` syntax check** across all adapters — pass
- **claude-code adapter dry-run** — pass, and it discovers `coco-loop`

## Scope note

This PR deliberately does **not** touch `README.md` or `skills/INDEX.md`. Those files hardcode skill counts, so per-skill PRs would all collide on them. Counts are reconciled in a single final PR once the series lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #50, opened 2026-07-31
chore(release): cut v1.2.0

coco-research:
Cuts **v1.2.0**. The last release, v1.1.0, shipped on 2026-07-18; everything since has landed on `main` without a release, so the published artifact is behind by two skills and a set of security fixes.

## What is in this release

**Added:** `coco-loop` (#45) and `scroll-world` (#46).

**Security**, all from @imachiever except the funding fix:
- Funding pointer moved off the reclaimed `rkz91` username (#47)
- Actions SHA-pinned to v7/v5, least-privilege `permissions`, DOM-XSS fix, CSP, CVE floor bumps (#48, was #37)
- Bootstrap commit-verification gate, Cursor adapter shell fixes, prompt-injection guidance (#49, was #38)

## Version bumps

`package.json`, `.claude-plugin.json`, `Formula/coco.rb` (url + version), and the README badge.

## Counts reconciled

This also closes out the count reconciliation that per-skill PRs deliberately deferred, so the numbers now match the tree exactly:

| | Before | After |
|---|---|---|
| Skills | 147 | **149** |
| Core skills | 64 | **66** |
| Addressable assets | 865 | **867** |
| Core install assets | 124 | **126** |

Verified against the tree: 66 core + 68 GSD + 6 Brain + 9 Super Intelligence = 149. `scripts/build-index.py` independently reports 140 skill directories, which plus the 9 Super Intelligence orchestration skills is the same 149. `coco-loop` and `scroll-world` were also inserted into the README core-skill catalog in alphabetical position.

## Two things to note

**The formula sha256 is a deliberate placeholder.** GitHub only generates the source tarball once the tag exists, so the checksum cannot be computed before then. It is set to zeros with a `# set after v1.2.0 tag` marker and will be filled by an immediate follow-up PR, exactly as #43 did for v1.1.0. **Do not publish the Homebrew tap until that lands.**

**Spec Version stays at 1.1.0.** That field tracks the framework spec rather than the release, and this version adds skills and fixes without changing any schema or contract. v1.1.0 bumped it only because of the relicensing. Say so if you would rather it track the release version and I will change it.

## Verification

`package.json` and `.claude-plugin.json` parse, `ruby -c Formula/coco.rb` reports Syntax OK, no stale count strings remain, and `check-command-refs.sh`, `check-evidence-gate.sh`, `install.sh` syntax, and the index-freshness check all pass locally.

### DIFF
--- a/.claude-plugin.json
+++ b/.claude-plugin.json
@@ -2,7 +2,7 @@
   "name": "coco",
   "displayName": "Coco",
   "description": "An entire team. Wherever your AI lives. 59 skills, 34 commands, 10 agents, 3 system bundles for project orchestration, knowledge tracking, and multi-agent pipelines. Open-core, vendor-neutral.",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "author": "Coco Inc",
   "license": "SEE LICENSE IN LICENSE",
   "homepage": "https://github.com/coco-research/coco",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,32 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) · Versioning: 
 
 ---
 
+## [1.2.0] — 2026-07-31
+
+### Added
+
+- **`coco-loop` skill.** Turns a plain-language goal into a readable charter that the person confirms, then arms and runs a bounded autonomous loop in propose-only mode. It never commits on its own. Requires the [coco-loops](https://github.com/coco-research/coco-loops) framework, which the skill now links with install instructions.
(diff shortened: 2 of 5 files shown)


rijul-mck (approved):
Approved. Version bumps are consistent across package.json, .claude-plugin.json, the formula and the README badge, and the counts now match the tree exactly (66 core + 68 GSD + 6 Brain + 9 SI = 149, independently corroborated by build-index.py reporting 140 directories plus the 9 orchestration skills). The formula sha256 placeholder is correct and expected at this stage; the follow-up PR sets it once the tag exists, so the Homebrew tap must not be published until then. Agreed on leaving Spec Version at 1.1.0 since no contract changed.

PR #51, opened 2026-07-31
chore(release): bump Spec Version to 1.2.0

coco-research:
Bumps the README **Spec Version** field from 1.1.0 to 1.2.0, per maintainer direction, so that it tracks the release version.

In #50 I had left it at 1.1.0 on the reasoning that v1.2.0 adds skills and security fixes without changing any schema or contract, and flagged the assumption for review. The maintainer's call is that Spec Version should follow the release version, so this corrects it before the tag is cut.

The v1.2.0 CHANGELOG section now records the bump under `### Changed`. The historical reference to the 1.0.0 to 1.1.0 bump in the 1.1.0 section is left untouched, since it accurately describes that release.

Landing this before tagging so that `v1.2.0` contains the correct value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@ Security contributions in this release are by [@imachiever](https://github.com/i
 
 - **Corrected asset counts** for the two new skills. Skills 147 to **149** (core 64 to **66**), total addressable assets 865 to **867**, core install 124 to **126** active assets.
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


rijul-mck (approved):
Approved. Spec Version now tracks the release version per maintainer direction, and the CHANGELOG records it. Landing before the tag so v1.2.0 carries the correct value.

PR #22, opened 2026-06-07
docs(readme): rewrite README to show true scale of 860 assets and 389-persona Superintelligence board

coco-research:
Rewrites the primary README to highlight the true scale of Coco (860 assets instead of 94), details the 9-team 389-persona Superintelligence advisory board, adds meta-orchestrator architecture design, and documents the core skills/GSD/Brain modules. Removed all unicode/emoji symbols.

PR #53, opened 2026-07-31
docs(readme): describe coco-loop and scroll-world, document the bootstrap install gate

coco-research:
Brings the README in line with what shipped in v1.2.0. Three real gaps, each verified against the tree rather than against the README's own claims.

**The remote bootstrap one-liner was undocumented.** `SECURITY.md` already refers to "the `bash <(curl -fsSL ...)` one-liner" as though the reader knows it exists, but the README never mentioned it. `git log -S"Quickest Install"` shows it was present before the rewrite in #22 and never came back after the confirmation gate landed in #49. There is now an "One-line remote install" row in the Installation Matrix giving the command, the confirmation-prompt behaviour, the `--yes` / `COCO_BOOTSTRAP_YES=1` bypass, and how to pass adapter and systems flags through it.

**The two new skills were invisible.** `coco-loop` and `scroll-world` appeared only inside the collapsed catalog block. Both now appear in the visible category slices with a short parenthetical.

**The security posture was scattered.** The confirmation gate, the pinned CI Actions, local-only execution and no telemetry each sat in disconnected places. One sentence after the Installation Matrix now ties them together and links `SECURITY.md`.

## A deliberate caveat on the pinning claim

Only `actions/setup-node` and `actions/setup-python` are pinned to full commit SHAs. `actions/checkout@v7` remains tag-pinned. The new wording names the two pinned actions rather than claiming blanket coverage, because the blanket claim would be false.

## Verified already correct, so left alone

Recounted independently against the tree: 149 skills (66 core, 68 GSD, 6 Brain, 9 Super Intelligence), 277 commands (35 core plus 242 generated), 34 agents, 389 personas across 9 departments and 70 cells, 15 rules, 3 workflows, 126 core-install assets, 867 total. All `--systems` and `--adapter` values, all internal links, and zero `rkz91` or `rijulkalra2000` references.

## Out of scope, reported not fixed

Three files carry stale counts and are worth a follow-up: `package.json` says 147 skills, `.claude-plugin.json` says "59 skills, 34 commands, 10 agents, 3 system bundles" which appears to date from v0.1.0, and `docs/getting-started.md` says "59 skills, 34 commands, 11 agents" and also omits the bootstrap one-liner.

## Gates

`check-command-refs.sh` PASS, `check-evidence-gate.sh` PASS, `build-index.py` regenerates with zero diff. Only `README.md` changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ A standard install equips your workspace with a lightweight core; full activatio
 CoCo ships **149 skills** (66 core + 83 across bundles). These are not prompt snippets — each is a full agent instruction set with state management, verification logic, and error handling. A representative slice by category:
 
 ### Visual design & styling
-`ui-ux-pro-max` (50 styles, 21 palettes, 50 font pairings, 9 stacks) · `frontend-design` · `design-taste-frontend` · `redesign-existing-projects` · `axiom-liquid-glass` (Apple Liquid Glass, WWDC 2025) · `swiftui-liquid-glass` · `web-design-guidelines` · `tailwind-patterns` (Tailwind v4) · `vercel-react-best-practices` · `clone-website` · `c4-architecture` · `arb-review` · `expo-api-routes` · `ai-product`.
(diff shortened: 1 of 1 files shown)


rijul-mck (approved):
Docs refresh: bootstrap install gate, coco-loop, scroll-world, security posture. CI green, README-only, no leaks. Approving.

PR #58, opened 2026-07-31
feat(skills): add HeyGen HyperFrames bundle (LICENSE + hyperframes skill, Apache-2.0)

coco-research:
First PR of the HeyGen HyperFrames bundle (`systems/hyperframes/`). Lands the shared bundle **LICENSE (Apache-2.0, Copyright 2026 HeyGen, Inc.)** + README + CREDITS entry, and the first skill `hyperframes`.

- Upstream: https://github.com/heygen-com/hyperframes · Apache-2.0 (verified via license API, raw LICENSE, npm manifest, marketplace.json) · npm `hyperframes` v0.7.87
- Placed at `systems/hyperframes/` with ONE LICENSE governing all bundle skills (follows the `systems/gsd/` convention). Not described as MIT anywhere.
- Bundle skills are not top-level, so they do not alter the generated `skills/INDEX.md`.

Gates green: leak scan (files+messages) · frontmatter · command-refs · evidence-gate · adapter dry-run · bash -n. Handoff Task 2 (HeyGen bundle), per-skill sequential.

PR #77, opened 2026-08-01
feat(skills): vendor embedded-captions skill (HeyGen HyperFrames bundle, Apache-2.0)

coco-research:
Final HeyGen bundle skill. Upstream ships a generated `fonts.css` base64-inlining 44 .woff2 faces; that generated artifact is **not committed** (tripped the repo leak gate; bloat). Instead the source `.woff2` binaries are kept, `fonts.css` is gitignored, and `inject-fonts.cjs`/`make-theme.cjs` build it on demand from the binaries (verified reproducible + self-heal tested). Governed by the bundle LICENSE (Apache-2.0, HeyGen) in #58. Gates green; no `--no-verify` used. Handoff Task 2.


rijul-mck (approved):
embedded-captions: fonts restructured to binaries (no base64 blob), self-heal verified, gates+CI green. Approving.

PR #78, opened 2026-08-01
Add Cognee knowledge graph memory bundle (rebased #36)

coco-research:
Rebased replacement for #36 (that branch could not be force-updated without re-scanning 41 already-published commits through the repo leak gate).

Adds `systems/cognee/` — three CoCo-authored skills (`cognee`, `cognee-store`, `cognee-recall`) wrapping Cognee's local HTTP API as an optional knowledge-graph memory backend alongside Brain. **No upstream Cognee code is vendored**; the skills are original write-ups.

**Upstream verified:** `topoteretes/cognee` — Apache-2.0, ~29.6k stars; `coco-research/cognee` confirmed a fork of it. Credited under *Reference material & SDKs* in CREDITS.md (was missing entirely).

**Two pre-merge fixes:**
- README.md left at main's version — the original PR's README edits predated v1.1.0/v1.2.0 and would have **regressed published claims** (Spec Version 1.0.0, "MIT" instead of open-core, 63 core skills). Bundle/skill count reconciliation happens in one dedicated pass, not in feature PRs.
- Genericized example payloads that hardcoded a real person's first name (`decided_by` / `DECIDED_BY` / `PARTICIPANTS`).

⚠️ **Locality caveat (worth stating plainly):** once a user actually runs Cognee, its *cognify* step needs an LLM and embedding provider, which commonly defaults to a cloud service. So "opt into Cognee" is **not** automatically as local as the Brain bundle, and that sits awkwardly beside CoCo's "100% local, no telemetry" positioning. Users should configure a local provider if they want to preserve that property.

Gates green: leak scan · build-index no-drift · frontmatter · command-refs · evidence-gate · adapter dry-run · bash -n.

### DIFF
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -37,6 +37,7 @@ Some skills teach or wrap third-party platforms. These are original write-ups bu
 | **Apple** — Liquid Glass design system (WWDC 2025) | `axiom-liquid-glass`, `swiftui-liquid-glass` | https://developer.apple.com |
 | **OpenAI** — Agents SDK, Realtime / Chat / Batch APIs, Whisper | `openai-agents`, `openai-api`, `openai-apps-mcp`, `openai-whisper` | https://platform.openai.com/docs |
 | **Microsoft** — Agent Lightning (RL agent training) | `agent-lightning` | https://github.com/microsoft/agent-lightning |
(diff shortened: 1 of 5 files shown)


rijul-mck (approved):
Cognee bundle, rebased. Upstream Apache-2.0 verified (topoteretes/cognee, fork confirmed), credit added, real name genericized, README left at main to avoid regressing published claims. Cloud-LLM caveat documented. Gates+CI green. Approving.

PR #82, opened 2026-08-02
fix(licence): remove Anthropic source-available skills and a docs scrape

coco-research:
## Why

Anthropic's [`anthropics/skills`](https://github.com/anthropics/skills) repository carries no repository-level licence; licensing is per-skill directory. Most of its skills are Apache-2.0, but four (`docx`, `pdf`, `pptx`, `xlsx`) ship a restrictive licence, and Anthropic's own README describes those four as "source-available, not open source".

That licence states that users may not:

- "Extract these materials from the Services or retain copies of these materials outside the Services"
- "Reproduce or copy these materials"
- "Create derivative works based on these materials"
- "Distribute, sublicense, or transfer these materials to any third party"

Publishing them in this repository performs at least three of those acts. Attribution cannot resolve this, because the restriction is on distribution itself rather than on credit. This is the same call already made for GitNexus (PolyForm-Noncommercial) and academic-research-skills (CC BY-NC 4.0).

## What was removed

| Path | Files | Reason |
|---|---|---|
| `skills/docx` | 57 | Verbatim copy of the upstream skill; already self-declared `license: Proprietary. LICENSE.txt has complete terms` |
| `skills/pdf` | 12 | Same |
| `skills/xlsx` | 1 | Upstream ships it under the same restrictive licence, but this repository carried it with **no** `LICENSE.txt` in any commit, with the upstream `license: Proprietary` frontmatter line deleted, and with a modified `SKILL.md` — so it also breached the no-derivative-works clause and left a reader no way to discover the restriction |
| `skills/writing-skills/anthropic-best-practices.md` | 1 (1,150 lines) | A verbatim scrape of Anthropic's public "Skill authoring best practices" documentation page, redistributed with no licence or attribution |

The scraped document is replaced in `skills/writing-skills/SKILL.md` with a link to the live page. That is better regardless of licensing, because the guidance stays current instead of frozen at the moment it was copied.

## What was deliberately NOT removed

`skills/skill-creator` and `skills/frontend-design` are Anthropic-published under **Apache-2.0** and are legitimately redistributable. Both `LICENSE.txt` files are intact and untouched. They do require attribution under Apache-2.0 section 4, which is handled in a separate pull request rather than here.

Every mention of `.docx`, `.xlsx` and `.pdf` as a **file extension** is also untouched — `commands/pm/sync-init.md`, `skills/doc-sync`, `skills/project-docs`, `skills/nfr-tracker`, `skills/meeting-notes`, and `systems/brain/.../document_reader.py` (whose `from docx import Document` refers to the unrelated `python-docx` library). Likewise `skills/skill-creator/SKILL.md` lines 186 and 313, where `docx` appears inside a fenced markdown block as illustrative example content.

Aggregate count claims in `README.md` ("Core skills (66)", "every one of the 149 skills") are also untouched. Both were already inaccurate before this change, and correcting them belongs to the separate index and counts work.

## Users who want these skills

`docs/recommended-plugins.md` already points at the Claude Code marketplace for `anthropic-skills (xlsx, docx, pdf, pptx)`, which is the licensed distribution channel.

## Verification

Every check below was run and its output read, not inferred from an exit code.

| Check | Result |
|---|---|
| `find . -name 'SKILL.md' -not -path './.git/*' \| wc -l` | 181 → 178 |
| `python3 scripts/build-index.py` | `Skills: 164 · Commands: 35 · Agents: 34` (164 is its own narrower glob, 167−3) |
| `bash tests/check-command-refs.sh` | `PASS: all command cross-references resolve.` |
| `bash tests/check-evidence-gate.sh` | `PASS: /team evidence-gate language present.` |
| `bash tests/smoke.sh` | `passed: 18  failed: 0` |
| Frontmatter lint (CI logic) | Checked 164 skill files, all valid |
| `install.sh --adapter {claude-code,codex,cursor,generic} --dry-run` | all exit 0 |
| Surviving skill links in `skills/INDEX.md` and `docs/by-domain/` | 0 |
| Leak scan | clean |

Indexes are regenerated in the same commit, so the "Verify INDEX is up to date" check passes.

## Note

This reflects a careful reading of the licence text present in the repository. It is not legal advice, and anything consequential should be confirmed by counsel. The removal reduces exposure either way, so it need not wait on that.

## Follow-up

Independent `docx`, `pdf` and `xlsx` skills may be written from the public library documentation (`python-docx`, `openpyxl`, `pdfplumber`, `pypdf`), copying nothing from the removed content. Whether to rewrite at all, versus pointing users at the official marketplace source, is an open decision.


rijul-mck (approved):
Approved. Verification evidence is captured in the PR body: every gate command was run and its output read rather than inferred from an exit code. CI is green on the required lint-and-test check.

PR #83, opened 2026-08-02
feat(team): architecture conformance gates — drift detection and built-as-declared

coco-research:
## Why

Every test suite shares one blind spot: **tests fail when behaviour breaks, not when architecture dissolves.** A component can be relocated, folded into another module, or deleted outright and a green build reports nothing. Requirements pass, coverage holds, and the agreed design is gone.

`/team` could previously satisfy every requirement and every test gate while having silently abandoned the module boundaries it was built against. Nothing in the pipeline could detect that. This closes the gap in both directions, enforced by Python scripts with real exit codes rather than by prose an agent can narrate around.

## What this adds

**Backwards — drift detection.** A commit-pinned index at `.arch/index.json` maps each architectural component to the real directories and files implementing it. `/team verify` gains failure mode **`(e)`, architecture abandoned**. Twelve validator checks; the central one is that every claimed path resolves on disk. A stale index pin degrades the gate to `UNVERIFIED`, never to clean.

**Forwards — built as declared.** `/team ship` Stage 1 records the architecture baseline without spawning agents, Stage 3 declares intended architecture to `.team-ship/ARCH-PLAN.json`, and the approval gate presents it as a table — the cheapest moment to disagree. Two gates then run alongside the seven test gates. A component declared and quietly dropped is a **BLOCK**. A file landing under a path the plan declared `outOfScope` is a **BLOCK**.

Path existence cannot be checked at Stage 3 because nothing is built yet, so the plan is written when the knowledge exists and tested when the evidence exists. Deferring the check rather than weakening it is what makes a forward declaration enforceable.

**New `/team arch` action:** `build` produces a fresh index, `drift` reconciles an existing one, `check` runs the validator alone with no model calls. `--no-arch` opts out and every gate then reports `DISABLED`, which is never a silent pass.

**Three roles** appended to `roles.md`: `solution-architect`, `architecture-reviewer`, `repo-cartographer`.

## Two fixes that came with it

- **The Cursor adapter never installed commands at all.** A Cursor user received zero slash commands while the distribution documentation advertised the full set. It now installs them, prunes unusable links, and reports skipped targets loudly. The Claude Code adapter gains the same end-of-run summary.
- **Seven toolkit assets existed but were invisible to every `/team` run** because nothing registered them. Also fixes a dangling `/pmstudio-prd` pointer.

## What it does not do

Stated here and in the artifacts themselves, because a gate that overstates its reach is worse than none. These are not hedges.

- **Structural drift only.** A component whose datastore was swapped *inside* its own already-claimed directory produces no drift. Every artifact says "structural drift only" in those words deliberately; please do not edit that wording out.
- **Not a work-allocation mechanism.** The index sits at C4-Container altitude, so one component can legitimately own thousands of files. Three separate design reviews rejected deriving per-agent file ownership from it, and `develop.md` states this explicitly.
- **Index currency is not mechanically enforced.** Comparing the pin to `HEAD` is prose an agent follows, not a script. This is the largest remaining correctness risk and the first thing worth instrumenting. Note that `.arch/pinned-commit` will read `STALE` immediately after this merges, until someone runs `/team arch drift`.
- **A clean result licenses no claim** about whether the code at those paths does what the component promised. That remains a review judgement.
- **Read budgets are prose-enforced.** There is no transcript API to count a subagent's tool calls from outside it, so the exploration read ceiling is guidance and no artifact reports it as verified.
- **One unverified assumption.** Cursor discovers user commands via `~/.cursor/commands/*.md`, which this repository's own `migrate-to-skills` skill documents, so commands install as `<namespace>:<command>.md` to match the Claude Code convention. Whether Cursor's parser is happy with a colon in the filename was not verified by launching Cursor. If not, the fix is a one-line separator change in `adapters/cursor/install.sh`.

## Verification

Every check run and its output read, not inferred from an exit code.

| Check | Result |
|---|---|
| `bash skills/arch-index/scripts/run_fixtures.sh` | 15/15 assertions, exit 0 |
| `validate_index.py .arch/index.json --repo-root .` | exit 0, 12/12 checks, 10 paths verified, 0 missing, 0 violations |
| `arch_drift.py --repo-root . --stdout` | runs clean, zero model calls |
| `bash tests/check-command-refs.sh` | `PASS: all command cross-references resolve.` |
| `bash tests/check-evidence-gate.sh` | `PASS: /team evidence-gate language present.` |
| `bash tests/smoke.sh` | `passed: 18  failed: 0` |
| Index freshness | `git status` clean after `build-index.py` |
| Leak scan | clean |

The three most load-bearing fixture assertions: `bad-paths.json` exits 1 naming the fabricated path; `plan-unbuilt.json` passes `--declare` but fails `--verify` naming both missing paths, and that asymmetry is the entire reason the forward gate works; a missing index or plan exits non-zero rather than being mistaken for a pass.

Only Python 3 standard library and `git`. No new dependencies.

## Attribution

Design adapted from [lak7/devildev](https://github.com/lak7/devildev), Apache-2.0. **No upstream source code was copied** — what was taken is prompt text, one schema idea and four procedures. Obligations under Apache-2.0 §4 are satisfied: a `CREDITS.md` row, an `## Attribution` section in `skills/arch-index/SKILL.md` with upstream file and line references, and licence footers in `references/tool-frugality.md` and `references/analysis-taxonomy.md` naming the upstream file, line range and modifications. Please do not remove those footers when reformatting. devildev ships no `NOTICE` file, so §4(d) imposes nothing further.

## Notes for review

`.arch/` is committed deliberately. An untracked index cannot be diffed across runs, would not exist in the clean checkout that `ship.md` Stage 11 verifies from, and would be silently erased by `git clean`, reverting every consumer to a degraded path without saying so. `.arch/DRIFT.json` and `.arch/tree.json` are derived scratch and are gitignored.

`.claude-plugin.json` is deliberately untouched: its counts are stale but they are being corrected from measurement in separate work, and half-correcting here would create a third inconsistent number. `skills/INDEX.md` and `commands/INDEX.md` are regenerated by the current generator, so their printed totals are that generator's narrow globs rather than the true tree count — also corrected separately.


rijul-mck (approved):
Approved. Verification evidence is captured in the PR body: every gate command was run and its output read rather than inferred from an exit code. CI is green on the required lint-and-test check.

PR #84, opened 2026-08-02
fix(index): make the generator count the whole tree and stop emitting dead links

coco-research:
## Why

`scripts/build-index.py` reported **165** skills where the tree holds **179**, and every count claim in the repository derives from that number. Worse, the CI step "Verify INDEX is up to date" enforced the wrong output as canonical — regenerating produced a clean tree, so the undercount looked correct.

This is upstream of every other counting problem in the repo. Fix the ruler before trusting any measurement it produced.

## Six defects, all in the generator

**1. Narrow globs.** `collect_skills()` globbed only `skills/*/SKILL.md` and `systems/*/skills/*/SKILL.md`. Two whole locations were invisible: `adapters/*/skills/*/SKILL.md` (5 skills) and `systems/*/*/SKILL.md` (9 superintelligence team skills, which have **no** `skills/` subdirectory). The four shapes now live in one `SKILL_SOURCES` table that declares each glob alongside the parent index naming its bundle, because the depths genuinely differ — and assuming one depth for all of them is what caused defect 4.

**2. Bundled skills were discarded.** `write_skills_index()` and `write_by_domain_views()` both contained `if 'bundle' in s: continue`, dropping all 111 bundled skills from every skill index. They are now grouped under `## Bundle: <name>` headings, reusing the shape `write_agents_index()` already used rather than inventing new grouping code.

**3. Links were built from the frontmatter name, not the directory.** 26 skills have a `name` that differs from their directory and 3 have none at all, producing **13 dead links** in `skills/INDEX.md`. Paths are now computed with `os.path.relpath` from the real file location, so they are correct from any index depth.

**4. The agents bundle was off by one.** `systems/<bundle>/agents/<name>.md` is three levels deep, so `parents[2]` resolved to the literal string `systems` — giving `agents/INDEX.md` a bundle named "systems" and **24 links** into a nonexistent `systems/systems/agents/` directory.

**5. `systems/INDEX.md` was promised and never written.** The module docstring advertised a bundle catalogue that did not exist, leaving bundles the least legible part of the repo. `write_systems_index()` now emits one — and it automatically flags any directory advertised as a bundle that installs no artifacts.

**6. A YAML parse error was silently swallowed.** `parse_frontmatter()` caught `yaml.YAMLError` and returned `None`, which is indistinguishable from "this file has no frontmatter". It now names the file and the error and exits non-zero. That immediately surfaced the real cause of a blank description row: `commands/util/architecture-review.md` had an unquoted `argument-hint` whose pipe characters YAML read as a block-scalar indicator. Quoting it restores the description.

## Verification

Every command run and its output read, not inferred from an exit code.

| Check | Before | After |
|---|---|---|
| `find . -name 'SKILL.md' -not -path './.git/*' \| wc -l` | 179 | 179 |
| `python3 scripts/build-index.py` reports | **165** | **179** |
| Dead links across all generated indexes | **37** (13 skills + 24 agents) | **0** of 220 |
| `systems/INDEX.md` | absent | present |
| Generator run twice | — | byte-identical output |
| `bash tests/check-command-refs.sh` | PASS | PASS |
| `bash tests/check-evidence-gate.sh` | PASS | PASS |
| `bash tests/smoke.sh` | 18/0 | 18/0 |
| `arch-index` fixtures | 15/15 | 15/15 |
| All four adapter dry-runs | exit 0 | exit 0 |

## What the new index shows

`skills/INDEX.md` now accounts for all 179 as **68 core + 111 across 6 bundles**. `systems/INDEX.md` renders the bundle table and emits:

> **Advertised as a bundle but installs no artifacts:** `team`

which surfaces a real finding automatically rather than requiring an audit to notice it.

One thing deliberately left visible rather than fixed: 61 of the 68 core skills group under **Unspecified**, because they lack the `domain` field that `CONTRIBUTING.md:47` and `docs/guides/migration.md:194` both declare **required**. That backfill is separate work. This change makes the gap legible instead of hiding it behind a filter.

## Not in scope

Stale count claims in `README.md`, `package.json`, `.claude-plugin.json` and `docs/` are untouched. They should be corrected from measurement once this generator is trustworthy, and correcting them in the same PR that changes the measuring tool would make the diff impossible to review.

### DIFF
--- a/agents/INDEX.md
+++ b/agents/INDEX.md
@@ -2,6 +2,8 @@
 
 Auto-generated. Run `python3 scripts/build-index.py` to refresh.
 
+**Total: 34 agents** — 10 core, 24 across 1 bundles.
+
 ## Core agents (10)
 
 | Agent | Description |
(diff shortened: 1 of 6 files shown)


rijul-mck (approved):
Approved. Before/after figures in the PR body were each produced by a command that was run and whose output was read: generator 165 to 179 matching find, dead links 37 to 0, output idempotent across two runs, all gates green.

PR #86, opened 2026-08-02
fix(si): remove internal project references and make the build scripts portable

coco-research:
Two problems, both in material public since 2026-06-07 (PR #18).

## 1. Internal project references — and a broken product promise

The Super Intelligence bundle anchored its persona evidence to an internal project by name: **260 references across 122 files.**

| What | Count |
|---|---|
| Citations of `/Users/user/Marvin/docs/architecture/*` as persona evidence | 34 |
| "Marvin Memory v2" project references | 113 |
| `marvin_v2_panel` JSON key in `registry.json` | 1 (+ its emitter) |
| `MARVIN_PG_TEST_DSN`, `MARVIN_TEST_POSTGRES_URL` in `commands/team/` | 2 |

**The second problem is the sharper one.** `README.md` sells the board as *"Citable and grounded — every stance is verified from public sources and carries a direct evidence URL."* Thirty-four of those evidence URLs pointed at a path on a private machine that no reader can open. They now read `internal synthesis (not publicly citable)` — honest about what a reader can and cannot verify, inside the one bundle sold as the differentiator.

Project references renamed to neutral terms: "Marvin Memory v2" → "founding Memory v2", "Marvin v2 panel" → "founding v2 panel", `marvin_v2_panel` → `founding_v2_panel` in both `registry.json` **and** the `build_registry.py` that emits it. One file renamed `05-marvin-v2-panel-attribution.md` → `05-v2-panel-attribution.md`, matching the convention every other persona already used; nothing referenced it.

### Deliberately preserved

**All six mentions of Marvin Minsky**, the MIT AI pioneer — an entirely unrelated, legitimate public reference. A blunt find-and-replace would have corrupted *"Won the Marvin Minsky Medal"* and his *Perceptrons* / *Society of Mind* entries. The replacements are ordered longest-first and his name is shielded during the pass.

### Deliberately unchanged

The `v2_panel_attribution` **frontmatter key**. It carries no project name, and both `build_registry.py` and the generated grounding-check prompts read it, so renaming it would break the registry build for no benefit.

## 2. Five build scripts that could never have worked

Every script under `systems/superintelligence/scripts/` set:

```bash
ROOT=/Users/user/projects/coco-platform/superintelligence
cd "$ROOT" || exit 1
```

That path exists for nobody, and it names a repository that no longer exists — so all five exited on the first line for every user, including the author. `ROOT` is now derived from the script's own location with an `SI_ROOT` override:

```bash
ROOT="${SI_ROOT:-$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)}"
```

Verified: `bash -n` passes on all five, and the derived value resolves to the directory containing `scripts/` and `ai/`.

One stale absolute path in `product-design/research/marty-cagan/notes.md` is now repo-relative.

## Method note

`registry.json` is generated from persona frontmatter, so I first checked whether regenerating was safe. It is **not** — `build_registry.py` produces a different file from the committed one (a `generated_at` bump plus content drift), so regenerating would have buried this change in unrelated diff. The registry was therefore edited directly and validated as JSON, and the emitter was updated to match so the two stay consistent.

## Verification

| Check | Result |
|---|---|
| References to the project name outside Marvin Minsky | **0** |
| Marvin Minsky mentions preserved | **6** |
| `/Users/user` paths remaining | **0** |
| `registry.json` parses | yes |
| `bash -n` on all five SI scripts | 5/5 |
| `build-index.py` | 179 skills |

## Scope

Git history retains the previous strings. That is a documented residual rather than a hidden one: the exposure is a project name with no credentials, client data or personal names, so rewriting public history was judged more disruptive than the risk warrants. Owner's decision.

Seven other `/Users/` hits were reviewed and deliberately left: example error output in `using-git-worktrees` and `systematic-debugging`, and text in `coco-ads`, `hyperframes-cli` and two GSD skills that instructs the reader to *redact* absolute paths. Those are correct as written.


rijul-mck (approved):
Approved. Zero remaining references outside the unrelated Marvin Minsky mentions, which were verified preserved. The five build scripts were confirmed broken before the fix and syntax-checked after.

PR #88, opened 2026-08-02
feat(m0): add M0 cross-tool agent memory as an opt-in bundle

coco-research:
## What M0 is

M0 answers one question well: **"where were we, what was verified, what is next."** It reads and writes a local operational thread so a session can be resumed from a different tool — Claude Code to Cursor and back — from a file, in milliseconds, with no daemon required.

Four skills, opt-in via `install.sh --systems m0`:

| Skill | Role |
|---|---|
| `m0` | control plane |
| `m0-remember` | write path |
| `m0-recall` | read path |
| `m0-handoff` | resume / handoff flow |

Plus a reference server and an MCP server — **Python standard library only, zero dependencies** — and `SPEC.md`, which is complete enough to reimplement the wire contract from without reading the code.

## Why the guarantees matter more than the feature

These are tested, not asserted:

- **Content-hash idempotency.** Re-posting an identical body returns the identical id *and* timestamp, so a caller can distinguish a retry from a new entry for free. That makes hooks and retries safe.
- **Sidecar spool on a locked store.** A write that cannot land is spooled atomically and reported `deferred: true` rather than lost, then drained on next start. This is exactly where naive memory writes get dropped: inside a shutdown hook.
- **Bounded timeouts.** Short by default and configurable. A two-minute timeout is indistinguishable from a hang, and `SPEC.md` says so explicitly rather than leaving it to taste.
- **Reads survive a locked store.** The read path unions the table with pending sidecars, so a concurrent bulk writer cannot black out recall.
- **Genuinely local.** SQLite on disk, loopback bind, no outbound calls, no telemetry. Delete the file and it has forgotten everything.

## What it does not do

`systems/m0/README.md` carries this as a side-by-side table so a reader can rule M0 out in about thirty seconds. There is no semantic search, no embeddings, no entity or relationship extraction, no graph to traverse, no summarisation or consolidation, no document ingestion, no automatic capture unless you install a hook, no authentication or real multi-tenancy, and no sync — one machine, one thread.

**Whenever the question is "what do we know about X" rather than "where were we", the bundled Cognee is the better choice, and the README says so directly.** Two memory bundles that overlapped without admitting it would be worse than one.

## Extraction note

This was extracted onto a fresh branch rather than rebased: the original work sat **35 commits behind** `main`. The change is purely additive apart from one documentation list.

## Also fixed here

`docs/INDEX.md` listed only GSD, Brain and Team under "By system bundle". It omitted **HyperFrames, Superintelligence and Cognee** — all of which install — and listed **Team, which is core rather than a bundle**. It now lists the six real bundles with skill counts and points at the generated `systems/INDEX.md` for live figures.

## Verification

Every check run and its output read.

| Check | Result |
|---|---|
| `bash systems/m0/tests/m0-smoke.sh` | **PASS 27, FAIL 0** against a live server |
| `python3 scripts/build-index.py` | 183 skills — matches `find` over the whole tree |
| `skills/INDEX.md` | gains a `## Bundle: m0 (4 skills)` heading |
| `systems/INDEX.md` | gains a row: `m0 · systems/m0/ · 4 skills` |
| Bundle gating, per skill | all 4 linked **with** `--systems m0`, **0** without |

**No installer change was needed** — `link_system()` already globs `systems/<name>/skills/*/`, which I confirmed by reading it rather than assuming.

## Honest caveats

Two things stated rather than buried. The Python `3.9+` floor is reasoned from the APIs used, not tested, because this machine has only 3.14. And `owner_user_id` / `visibility` are columns with defaults rather than enforcement — `SPEC.md` and the README both say so explicitly instead of implying a security property that is not there.

### DIFF
--- a/docs/INDEX.md
+++ b/docs/INDEX.md
@@ -21,9 +21,19 @@
 
 ## By system bundle
 
-- [GSD](../systems/gsd/) — project orchestration (68 skills)
+Installable with `install.sh --systems <name>`:
+
+- [GSD](../systems/gsd/) — project orchestration (68 skills, 24 agents)
+- [HyperFrames](../systems/hyperframes/) — video and motion suite (20 skills)
+- [Superintelligence](../systems/superintelligence/) — 389-persona expert board (9 team skills; generates 242 commands at install)
 - [Brain](../systems/brain/) — local knowledge tracker (6 skills)
-- [Team](../systems/team/) — multi-agent pipelines
+- [M0](../systems/m0/) — cross-tool agent memory (4 skills)
+- [Cognee](../systems/cognee/) — knowledge-graph memory (3 skills)
+
+See [systems/INDEX.md](../systems/INDEX.md) for the generated bundle table with live counts.
+
+`/team` is **not** a bundle — it is core, and its commands are installed by a plain
+`bash install.sh`. See [systems/team/](../systems/team/) for the design notes.
 
 ## By IDE adapter
 
--- a/skills/INDEX.md
+++ b/skills/INDEX.md
@@ -2,7 +2,7 @@
 
 Auto-generated. Run `python3 scripts/build-index.py` to refresh.
 
-**Total: 179 skills** — 68 core, 111 across 6 bundles.
+**Total: 183 skills** — 68 core, 115 across 7 bundles.
 
 ## Design (14)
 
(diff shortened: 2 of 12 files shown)


rijul-mck (approved):
Approved. The 27-assertion smoke suite was re-run against a live server rather than taken on report, and bundle gating was verified per skill: all four present with the flag, none without. The README's non-capabilities list is unusually honest and should stay that way.

PR #89, opened 2026-08-03
fix: close the freshness gaps that let tonight's own counts go stale again

coco-research:
## Why

Seven pull requests merged earlier tonight (#82–#88) corrected every documented count in the repository. **Within two of those pull requests, three counts were stale again.**

That is the real defect, and it is missing enforcement rather than a typo. Nothing checks these claims, so they rot the moment anything changes. This PR fixes the enforcement first and the symptoms second.

## Root causes — two gates added to `ci.yml`

**`systems/INDEX.md` was generated by #84 and then omitted from the freshness check.** It appears in neither the output-existence loop nor the "Verify INDEX is up to date" diff, so it could drift silently for as long as nobody happened to regenerate and look. It was in fact **already stale** — this PR's own commit shows it changing by one line. Now covered by both.

**A bundle advertised through `--systems` was never exercised in CI.** Only a bare `--dry-run` ran per adapter, which is precisely how `--systems team` — a bundle that installs nothing — survived unnoticed. There is now a gate that reads each manifest's `supports_systems` and requires every advertised bundle to actually change that adapter's install plan. An empty list is a checked claim too: the adapter must either reject the flag or leave the plan provably unchanged.

That makes the manifests **enforceable rather than decorative**. Over-claiming a bundle now fails the build.

**Three suites that pass locally ran nowhere in CI:** `tests/smoke.sh`, `skills/arch-index/scripts/run_fixtures.sh`, and `systems/m0/tests/m0-smoke.sh`. All three now run, and the step also fails if a suite leaves the tree dirty.

## Stale claims corrected

| File | Was | Now |
|---|---|---|
| `skills/README.md` | a `Documents \| 3` row for `docx`/`pdf`/`xlsx`, removed in #82 — table summed to 30 against 27 real entries | row deleted |
| `.claude-plugin.json` | advertised full-install totals while its hook passes no `--systems` | states the measured **68 skills, 37 commands**, marks bundles opt-in, keeps full-install figures as what enabling every bundle reaches |
| three adapter manifests | `m0` published as a bundle in #88 but listed in none of them | `m0` added |
| `systems/hyperframes/README.md` | "one LICENSE governs every file"; "all 20 skills from one upstream under one license" | corrected, see below |
| `agents/README.md` | silent on why `find` returns 37 agents while 34 install | documents the convention |

### The HyperFrames licence claim was wrong in a way that matters

That README asserted a single Apache-2.0 licence covers everything beneath it. It does not:

- `skills/talking-head-recut/` carries its own `NOTICE.md` — MIT, copyright leeoxiang, from `notedit/vtake-skills`, not HeyGen
- two vendored **GSAP** builds are under GreenSock's own Standard License
- **57 bundled font files** (Hershey stroke fonts, Fontsource redistributions of Google Fonts) carry their originating licences; 14 files in the bundle reference GreenSock, Hershey or Fontsource

The section now states this and that a nearer notice wins.

### The 37-vs-34 agent discrepancy is a convention, not a bug

The three extra files are `builder.md`, `director.md`, `finalize.md` under `systems/hyperframes/skills/motion-graphics/agents/`. They have no frontmatter and are not role definitions — they are prompt bodies dispatched at named stages of that skill's own pipeline, listed in the `prompt` column of its stage table. Internal to one skill, so deliberately neither linked nor indexed. `agents/README.md` now says so, and separates the three role populations that are easy to confuse.

## Two things this PR deliberately does not assert

**No count is quoted for the `/team` roles.** Three counting methods over `commands/team/roles.md` gave 40, 34 and 44, depending on whether the `## Role Template` placeholder and each role's `### System Prompt` subheading are counted. The file is cited instead. Quoting a fourth guess would be the same defect this PR exists to fix.

**`adapters/generic/manifest.json` keeps all six bundles.** I removed them mid-work on the evidence that generic parses no `--systems` flag of its own, and the new gate immediately caught that as wrong: generic delegates to the shared renderer and its install plan does change per bundle. Restored. The gate catching my own bad edit is the clearest argument for having it.

## Verification

| Check | Result |
|---|---|
| Bundle gate, **extracted verbatim from `ci.yml`** and run under `bash` | **20 / 20 PASS** — 6 bundles × claude-code, codex, generic, plus both no-bundle adapters |
| `systems/m0/tests/m0-smoke.sh` | PASS 27, FAIL 0 |
| `build-index.py` | 183 skills, matching `find` |
| `ci.yml` parses as YAML; every touched JSON valid | yes |

The gate was extracted from the workflow rather than paraphrased, because a hand-written approximation of a CI step is not evidence that the CI step works. It also revealed that my first version of the loop tested only the first bundle: **zsh does not word-split unquoted expansions and bash does**, so a maintainer running it by hand would have seen a false pass. It now reads a newline-delimited list with `while read`.

### DIFF
--- a/.claude-plugin.json
+++ b/.claude-plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "coco",
   "displayName": "Coco",
-  "description": "An entire team. Wherever your AI lives. 179 skills, 37 commands, 34 agents, 5 system bundles for project orchestration, knowledge tracking, and multi-agent pipelines. Open-core, vendor-neutral.",
+  "description": "An entire team. Wherever your AI lives. This install delivers 68 skills, 37 commands and 10 agents. System bundles are opt-in via `install.sh --systems <name>`, and enabling every bundle reaches 179 skills, 279 commands and 34 agents for project orchestration, knowledge tracking, and multi-agent pipelines. Open-core, vendor-neutral.",
   "version": "1.2.0",
   "author": "Coco Inc",
   "license": "SEE LICENSE IN LICENSE",
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,7 +87,7 @@ jobs:
         run: |
           python3 scripts/build-index.py
           # confirm all expected outputs exist
-          for f in skills/INDEX.md commands/INDEX.md agents/INDEX.md docs/by-domain/pm.md docs/by-domain/engineering.md; do
+          for f in skills/INDEX.md commands/INDEX.md agents/INDEX.md systems/INDEX.md docs/by-domain/pm.md docs/by-domain/engineering.md; do
             test -f "$f" || { echo "Missing: $f"; exit 1; }
           done
           echo "INDEX files generated successfully."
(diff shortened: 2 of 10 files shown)


rijul-mck (approved):
Approved. The bundle gate was extracted verbatim from ci.yml and run under bash rather than paraphrased, giving 20/20. It already caught one bad edit in this PR's own development, which is the strongest evidence it works.

PR #90, opened 2026-08-03
docs: name the HyperFrames licence and scope precisely in CREDITS.md

coco-research:
Closes the last attribution gap flagged by the requeue audit. This was marked attended because the propose-only loop is forbidden from editing `CREDITS.md`, which is correct — it is legal attribution text under Apache-2.0 §4.

## What was wrong

The HyperFrames row was imprecise in three ways simultaneously:

| Field | Was | Now |
|---|---|---|
| Licence | `See upstream` | `Apache-2.0 (Copyright 2026 HeyGen, Inc.)` — from `systems/hyperframes/LICENSE:178` |
| Upstream | `npmjs.com/package/hyperframes` — the only npm-pointing row in the file | `github.com/heygen-com/hyperframes` |
| What CoCo took | only "the HTML-composition + local render engine `coco-ads` drives" | the **whole 20-skill bundle** vendored at `systems/hyperframes/` |

The third was the substantive one. The repository vendors 20 skills, and the row credited a single engine.

## The caveat it now carries

Apache-2.0 does **not** govern every file under that directory, and anyone reusing an individual file needs to know:

- `skills/talking-head-recut/` has its own `NOTICE.md` — MIT, leeoxiang, from `notedit/vtake-skills`
- two vendored **GSAP** builds are under GreenSock's own licence
- **57 bundled font files** carry their originating font licences

The row points at `systems/hyperframes/README.md` for the breakdown rather than duplicating it, so there is one place to keep current. That README was corrected in #89.

## Verification

| Check | Result |
|---|---|
| Row cites repo, exact licence, copyright holder, vendored path | all four present |
| `grep -c npmjs CREDITS.md` | **0** |
| `See upstream` on this row | gone |
| `bash tests/check-command-refs.sh` / `check-evidence-gate.sh` | PASS |
| Files changed | exactly 1 |


rijul-mck (approved):
Approved. Licence and holder taken from systems/hyperframes/LICENSE:178, and the sub-vendor caveat points at the README corrected in #89 rather than duplicating it.

PR #85, opened 2026-08-02
docs: reconcile every count from measurement, and close the attribution gaps

coco-research:
Two commits, reviewable independently.

## 1. Counts corrected from measurement

With the index generator fixed (#84), every documented count could finally be checked against the filesystem. Eleven documents carried figures that disagreed with each other and with reality, and **not one stated the true skill total**.

Every number was re-derived, not adjusted:

| Quantity | Command | Value |
|---|---|---|
| Skills | `find . -name 'SKILL.md' -not -path './.git/*' \| wc -l` | **179** |
| Core skills | `ls -d skills/*/ \| wc -l` | 68 |
| Commands in-repo | `find commands -name '*.md' ! -name INDEX.md \| wc -l` | 37 |
| Commands generated at install | `build_commands.py` + `build_meta_commands.py` | 242 |
| Agents installed | `agents/` + `systems/*/agents/` | 34 |
| Personas | `systems/superintelligence/**/personas/*.md` | 389 |
| Rules | `rules/cursor-mdc/*.mdc` | 15 |

Corrected in `package.json`, `.claude-plugin.json`, `README.md` (six places, including the asset table and the "867 total" that never added up — it is 896), `docs/getting-started.md`, `docs/INDEX.md`, `docs/install.md`, `docs/distribution/cursor-extension.md`, `docs/distribution/claude-code-marketplace.md`, `docs/guides/comparison.md`, and a `CHANGELOG` entry that described a correction which had not actually landed.

### Two claims were backwards, not merely stale

**`docs/install.md` said `--systems` works on "claude-code, cursor only".** Measured per adapter with `--systems gsd --dry-run`:

| Adapter | Result |
|---|---|
| claude-code | exit 0, acts on it |
| codex | exit 0, acts on it |
| generic | exit 0, acts on it |
| **cursor** | **exit 1, `Unknown flag: --systems`** |

The sentence named the one adapter that rejects the flag and omitted the two that support it.

**Every adapter manifest declared `supports_systems: ["gsd","brain","team"]`.** That listed `team`, which installs nothing; omitted `cognee`, `hyperframes` and `superintelligence`, which all install; and asserted Cursor support that does not exist. Each manifest now declares what its adapter actually accepts, with `cursor` declaring an empty list. Note that **no code reads this field today** — it is corrected so it stops actively misleading a reader, not because it changes behaviour.

### `/team` is core, and now says so

[...]

PR #91, opened 2026-08-04
fix(cli): stop advertising the phantom team bundle in the CLI's own help

coco-research:
`bin/coco.js` offered `--systems gsd,brain,team` in two places — its usage comment (line 8) and its `--help` output (line 106).

`team` installs nothing. `systems/team/` holds documentation only, and passing it to `--systems` has no effect. That claim was corrected across the README, the docs tree and all five adapter manifests in #85 and #89 — **but the CLI was missed**, so the one surface a new user reads first still recommended it.

The flag now lists the six bundles that actually install: `gsd`, `brain`, `cognee`, `hyperframes`, `m0`, `superintelligence`. That set comes from `supports_systems` in `adapters/claude-code/manifest.json` rather than being written by hand, and #89's bundle gate verifies every entry in that field genuinely changes the install plan — so the two cannot silently drift apart again.

## Why this is worth a PR rather than a footnote

The package is on the verge of first publication to npm. `package.json` is fully configured (`@coco-research/coco-cli@1.2.0`, `bin`, `files`, `publishConfig: public`), `README.md:541` advertises `npm install -g @coco-research/coco-cli`, and `skills/coco-cli` is a shipped skill that wraps it. Publishing today would ship a CLI whose `--help` recommends a no-op flag.

## Verification

| Check | Result |
|---|---|
| `grep ',team' bin/coco.js` | no match |
| `node --check bin/coco.js` | passes |
| `node bin/coco.js --help` | exit 0, renders the corrected list |
| List matches `supports_systems` | exact |
| `bash tests/smoke.sh` | passes |
| Files changed | 1 |

### DIFF
--- a/bin/coco.js
+++ b/bin/coco.js
@@ -5,7 +5,7 @@
 //   npx @coco-research/coco-cli               # clones to ./coco and installs (auto-detect adapter)
 //   npx @coco-research/coco-cli install       # same as above
 //   npx @coco-research/coco-cli install --adapter cursor
-//   npx @coco-research/coco-cli install --systems gsd,brain,team
+//   npx @coco-research/coco-cli install --systems gsd,brain,m0
 //   npx @coco-research/coco-cli update        # pull latest in existing clone
 //   npx @coco-research/coco-cli uninstall     # remove symlinks + clone
 //   npx @coco-research/coco-cli --help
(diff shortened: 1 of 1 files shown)


rijul-mck (approved):
Approved. Bundle list now derives from supports_systems, which #89's gate verifies, so it cannot drift again. Worth landing before first npm publish.

PR #92, opened 2026-08-04
feat(npm): rename the package to cocosuperintelligence and publish unscoped

coco-research:
The package was named `@coco-research/coco-cli` and **had never been published**, so there is no installed copy of the old name anywhere and nothing downstream breaks. The owner's chosen name is `cocosuperintelligence` — it matches how the README presents the product, and it shortens the install:

```bash
npx cocosuperintelligence
```

## Scope of the rename

46 occurrences across 8 files: `package.json`, `bin/coco.js`, `README.md`, `install.sh`, `skills/coco-cli/SKILL.md`, `skills/cli-anything/SKILL.md`, `docs/distribution/npm-github-packages.md`, `.github/workflows/publish-npmjs.yml`. Generated indexes regenerated so the catalog descriptions follow.

`publishConfig.access` was **dropped** rather than updated — it only applies to scoped packages, and an unscoped package is public by default, so keeping it would imply a constraint that doesn't exist.

## Two deliberate non-changes

**The dated `CHANGELOG` entries that mention the old scoped name are left as written.** They record what was true when written; rewriting them would make the record dishonest. A new `[Unreleased]` entry documents the rename instead.

**`skills/coco-cli/` keeps its directory name.** It describes what the skill does — wrap the CLI — and renaming it would churn every generated index and any link pointing at it for no user-visible gain. Its contents now reference the new command.

## Verification

| Check | Result |
|---|---|
| `node bin/coco.js --help` | exit 0; every usage line reads `npx cocosuperintelligence` |
| `package.json` | parses; `name: cocosuperintelligence` |
| `publish-npmjs.yml` | valid YAML |
| `bash tests/check-command-refs.sh` / `check-evidence-gate.sh` | PASS |
| `npm pack --dry-run` | `cocosuperintelligence-1.2.0.tgz`, 19.3 kB, same 5 files |
| Old name outside CHANGELOG history | zero references |

### DIFF
--- a/.github/workflows/publish-npmjs.yml
+++ b/.github/workflows/publish-npmjs.yml
@@ -1,7 +1,7 @@
 name: Publish to npmjs.com
 
-# Publishes @coco-research/coco-cli to the public npmjs.com registry.
-# Frictionless install for users — no auth required: `npm install -g @coco-research/coco-cli`.
+# Publishes cocosuperintelligence to the public npmjs.com registry.
+# Frictionless install for users — no auth required: `npm install -g cocosuperintelligence`.
 #
 # Setup (one-time):
 #   1. Create npm account at https://www.npmjs.com/signup
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) · Versioning: 
 
 ## [Unreleased]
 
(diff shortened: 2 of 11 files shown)


rijul-mck (approved):
Approved. Package was never published so the rename breaks nothing. CLI help verified to read npx cocosuperintelligence throughout, and the tarball is unchanged at 5 files / 19.3 kB.

PR #95, opened 2026-08-14
feat(skills): add coco-diagram editorial diagram skill

rijul-mck:
PR 1 of 5 for the diagram capability.

Adds `skills/coco-diagram` — editorial diagram system emitting self-contained HTML + inline SVG. 27 visual types, semantic patterns, draw.io / Mermaid **redraw** (not render), accessible-SVG contract, format/size/detail/audience dials.

**Provenance, stated plainly:** adapted from [diagram-design](https://github.com/cathrynlavery/diagram-design) by Cathryn Lavery, **MIT**, pinned at upstream **v2.3.2**. Upstream's MIT notice is retained verbatim at `skills/coco-diagram/LICENSE` — MIT requires the notice to travel with substantial portions — plus a `CREDITS.md` row.

**Rebranded to CoCo:**
- skill renamed `diagram-design` → `coco-diagram`
- accent token: upstream orange → **CoCo violet `#8B5CF6`**, incl. all rgba tints and the dark-mode variant `#A78BFA` (433 hex + ~180 rgba occurrences)
- the accent role was named after its color upstream; renamed to the semantic `accent` throughout (206 occurrences) so a future brand change doesn't invalidate the prose
- consultancy names removed from the quadrant copy
- accent no longer matches upstream's default, which also satisfies the skill's own §0 first-run style-guide gate

**Not vendored:** upstream dev tooling and `docs/screenshots` (3.7 MB). Shipped payload is `SKILL.md` + `references/` + `assets/` + 3 skill-local scripts (**145 files, 2.1 MB**). The 95 example HTML files are kept because `references/` links to 103 of those paths across 31 files.

Gates green: leak scan · index no-drift · frontmatter · command-refs · evidence-gate · adapter dry-run · `bash -n`. `skills/INDEX.md` regenerated (184 skills).

Follow-ups: PR 2 commands, PR 3 mermaid-overlap, PR 4 counts, PR 5 changelog.

PR #96, opened 2026-08-14
fix(coco-diagram): WCAG accent, dangling citations, unshipped-script instructions, icon attribution

rijul-mck:
Five defects found by reviewing #95. **Three were introduced by #95.**

### 1. WCAG AA failure on text — introduced by #95
The rebrand accent `#8B5CF6` on paper `#f5f5f5` is **3.88:1**; text AA needs 4.5:1. 97 `<text>` elements affected, and the large-text 3:1 exemption does not apply — examples use `viewBox="0 0 1000 …"` at `width: 900px`, so a nominal 12px label renders ~10.8px.

Darkening to violet-700 `#6D28D9` fixed contrast (6.52:1) but **broke colourblind separation from the `link` role**: deuteranopia distance fell 104 → 41, protanopia 95 → 31, against a ~60 floor. Violet's signal against blue *is* its red channel, which deuteranopia suppresses.

Final accent is rose-700 **`#BE185D`** / dark **`#F472B6`**, which clears both gates:

| | text AA on paper | deuteranopia vs link | protanopia |
|---|---|---|---|
| `#8B5CF6` (shipped in #95) | 3.88:1 ❌ | 104 | 95 |
| `#6D28D9` (first attempt) | 6.52:1 | **41** ❌ | **31** ❌ |
| `#BE185D` (this PR) | **5.54:1** ✅ | **126** ✅ | **104** ✅ |

**Root cause fixed too:** `style-guide.md` specified a contrast contract for `ink` and `muted` on `paper` but none for `accent` — nothing could catch the regression. That clause plus a runnable luminance check are now in the file.

### 2. Eleven dangling asset citations
`example-*-extended*.html` was cited 12 times across 4 reference files. I verified those files exist **neither here nor upstream** (`find` for `*extended*` in the upstream tree → 0 hits), so vendoring them is impossible. Citations removed; `type-process.md` §12's heading no longer names a nonexistent file.

### 3. Six instructions to run unshipped scripts
`animation.md` had an unhedged `Run:` block for three scripts we do not vendor; `SKILL.md:286` a fourth. Two more hedged with *"In this repository"* — true upstream, false here. All six now point at the shipped `self_check.py` and label the rest upstream-only.

### 4. Icon attribution incomplete
The skill ships **87 vendored icon paths** from Tabler (MIT), Simple Icons (CC0), log-z/logos (MIT), Devicon (MIT). Inline credits named the projects, but no licence text travelled with the copy. New `THIRD-PARTY-LICENSES.md` records each source, licence, upstream, licence-text location, the trademark note, and the fact that upstream declares **no licence** for the Stata mark. (Upstream MIT year verified correct at 2025 — not fabricated.)

### 5. Missing `domain` — introduced by #95
It was the only one of 69 core skills with no `domain`, so it landed in an `## Unspecified` INDEX section and was excluded from every `docs/by-domain/` view. Now `domain: design`; Unspecified section is gone, Design is 15.

Also: H1 retitled to *CoCo Diagram*, version pinned `2.3` → `2.3.2`, and three in-skill command names moved to the `/diagram` namespace, dropping a false claim that a command file lives at the repo root.

**Verification:** all 95 examples + 5 templates pass `self_check.py` after the 436-site palette change (100/100). Gates green: command-refs, evidence-gate, adapter dry-run, `bash -n`, index regenerated.


coco-research (approved):
Fixes three defects #95 introduced (WCAG AA text contrast, missing domain field, and — after the first fix attempt — colourblind separation from the link role) plus two inherited (11 dangling asset citations, 6 instructions to run unvendored scripts), and closes the icon attribution gap with THIRD-PARTY-LICENSES.md. Contrast and colourblind distances independently recomputed; 100/100 assets pass self_check after the 436-site palette change. Approving.

PR #100, opened 2026-08-18
feat(site): add the launch film to the homepage flagship band

rijul-mck:
## What this adds

A "Watch the film (20s)" link in the flagship band's CTA row, opening the launch video in a centered lightbox — the same "Watch the film" pattern Apple uses next to its own hero content, rather than autoplaying it inline or replacing the existing flagship copy (which the previous PR deliberately trimmed to a single claim).

## The video

20 seconds, built with the `/coco-ads` skill and HyperFrames, rendered fully locally (no cloud render path, no vision API). It embeds two real product recordings — the install confirmation prompt and a live `/SI-Decide` run — rather than recreating either as fabricated UI.

## Implementation

- Lightbox reuses the exact open/close/backdrop/Escape pattern already shipped for the products overlay in the previous PR, for consistency.
- Respects `prefers-reduced-motion`: doesn't auto-play when that's set, though the video still exposes native controls either way so it's always watchable.
- Poster frame extracted from the video itself (not a separately designed asset) so the lightbox never shows a blank/black frame before playback starts.

## Verification

- All repository gates pass: `scripts/build-index.py` (no drift), `tests/check-command-refs.sh`, `tests/check-evidence-gate.sh`.
- All 7 asset references on the homepage resolve against disk.
- Tag balance checked.
- Lightbox open/close/video-pause-on-close behavior verified in-browser (computed styles, video play state) before this PR.


coco-research (approved):
Approved. CI passes, all asset refs resolve, and the lightbox open/close/pause behavior was verified in-browser before this PR.

PR #103, opened 2026-08-19
fix(site): taste-skill sweep (hero density, mobile nav, em-dashes, motion speed) + CoCo Code copy fix

coco-research:
## What this does

Closes out items 1-10 of a design-taste audit run this session (a structured checklist pass, verified against the live site with real browser measurements, not just source reading), plus a direct correction to CoCo Code's product copy.

- **`/coco/`'s H1 line wrap** — verified live that it wraps to 3 lines instead of 2 specifically in the 861-1023px gap between the mobile breakpoint and full desktop. Fixed with one intermediate font-size step, applied to both pages.
- **Hero density** — both heroes trimmed from 5 text blocks to the intended 4. Nothing was cut, just relocated to where it fits: the "regulated-risk product leader" line moved into the homepage's About section; "runs on the model you already have" moved into `/coco/`'s Install section, next to the install command it's actually about.
- **Mobile nav** — fixed every plain nav link disappearing below 860px with zero replacement. Every destination is now reachable via horizontal scroll (the same pattern already used for the product tabs).
- **CTA dedup** — the homepage's "go to `/coco/`" link repeated three times in quick succession; removed the closest, most redundant one.
- **Em-dashes** — all 9 present in actual visible copy fixed (title tag, hero subtext, a product card, a credit line, a stage description). Precisely distinguished from em-dashes inside CSS/JS *source comments*, which no visitor ever sees and were left alone.
- **Tap targets** — secondary "X ›" links widened; they sat right at the WCAG 2.2 AA floor with zero padding.
- **Motion speed** — the reveal-on-scroll animation was flagged directly as feeling slow while scrolling. Cut from 0.8s to 0.45s, 24px to 14px of travel, tighter stagger. Same mechanism, same `prefers-reduced-motion` handling, just faster.
- **GitHub CTA verb** — unified three different verbs for the same link ("Get it" / "See it" / "Get CoCo") to one.
- **CoCo Code copy** — was stale, describing a small fork "headed toward its own project." The current, accurate description came directly from the project owner and was independently verified against the real local project before publishing: the Unix-socket protocol, the `coco`/`coco-app` binary names, the zero-egress model, and the still-present jcode MIT license notice all checked out against the project's own README and docs. Also corrected the display name to "CoCo Code" (matching the project's own README) from the site's prior "CoCoCode".

## Verification

- All repository gates pass: `scripts/build-index.py` (no drift), `tests/check-command-refs.sh`, `tests/check-evidence-gate.sh`.
- Tag balance checked on both pages.
- Em-dash sweep re-verified with comment-state tracking to avoid false positives from CSS/JS comments.
- All 7 asset references per page resolve against disk.
- Hero, mobile-nav, and CoCo-Code-copy fixes spot-checked live in-browser (computed styles, DOM content, screenshots) before this PR.

### DIFF
--- a/coco/index.html
+++ b/coco/index.html
@@ -104,7 +104,7 @@ section.tint{background:var(--surface)}
 .btn{display:inline-block;background:var(--blue);color:#fff;font-size:17px;letter-spacing:-.022em;
   padding:12px 24px;border-radius:var(--stadium);transition:opacity var(--dur) var(--ease)}
 .btn:hover{opacity:.85;text-decoration:none}
-.arrow{font-size:17px;letter-spacing:-.022em}
+.arrow{font-size:17px;letter-spacing:-.022em;display:inline-block;padding:10px 0;margin:-10px 0}
 .arrow:after{content:" ›";font-weight:600}
 
 /* ---- hero visual: the 9 departments. Pure CSS/SVG, no imagery. ---- */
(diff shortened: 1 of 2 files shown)


rijul-mck (approved):
Approved. CI passes, gates clean, hero/nav/copy fixes verified live in-browser before this PR.

PR #106, opened 2026-08-19
feat(site): add meta description + OG/Twitter tags, fix About nav label

coco-research:
## Summary
- TODO item 17: both pages had only `charset`/`viewport`/`title` in `<head>` — shared links unfurled with no description or image. Added `meta description`, Open Graph, and Twitter Card tags to both pages, reusing each page's own existing verified copy (no new claims), pointing at the `og-image.png` already live in the repo at `/assets/og-image.png`.
- TODO item 18: renamed the home nav's "About" link to "Principles" — it points to the "What stays the same across every product" section, which never contained the word "About."

## Test plan
- [x] Ran `scripts/build-index.py`, `tests/check-command-refs.sh`, `tests/check-evidence-gate.sh` — all pass
- [x] Verified in browser preview: meta tags present and correct on both pages, `og:image` resolves to the absolute root-level image (important since `/coco/`'s own relative asset paths point at a different local `assets/` dir), nav renamed and pointing at the same anchor

### DIFF
--- a/coco/index.html
+++ b/coco/index.html
@@ -4,6 +4,16 @@
 <meta charset="utf-8">
 <meta name="viewport" content="width=device-width, initial-scale=1">
 <title>CoCo: An entire engineering department, inside your editor</title>
+<meta name="description" content="Each of the 389 requires a live, checkable source. Then 184 skills ship what they decided.">
+<meta property="og:type" content="website">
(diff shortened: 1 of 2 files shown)


rijul-mck (approved):
Gates pass, CI green, meta tags and nav rename verified in preview on both pages.

PR #107, opened 2026-08-19
feat(adapters): real VS Code adapter, replacing the Continue stub

### LINKED ISSUE #4: Add adapter for VS Code (via Continue)
## Goal
Implement the planned VS Code adapter that wires Coco artifacts via the Continue extension's config format.

## Files involved
- `adapters/vscode-continue/install.sh` (new)
- `adapters/vscode-continue/manifest.json` (new)
- `adapters/vscode-continue/README.md` (new)

## Acceptance criteria
- [ ] Reads Coco's skills/commands and translates to Continue's config schema
- [ ] Writes to `~/.continue/config.json` (or appropriate path)
- [ ] Idempotent on re-run
- [ ] `--dry-run` flag works
- [ ] Documented in README's compatibility table

## Hints
- Continue config format: https://docs.continue.dev/customize/config
- Reference: existing `adapters/cursor/install.sh` and `adapters/codex/install.sh`
- This is planned for v0.2 — you'd be helping unblock that release

rijul-mck:
## What this fixes

Coco installed cleanly into Claude Code and Cursor, but VS Code got a stub that told you to install the Continue extension. That stub was written on the premise that VS Code has no skills system of its own — which is no longer true.

VS Code discovers agent skills, agent files and instruction files under `~/.copilot/{skills,agents,instructions}` by default (`chat.agentSkillsLocations`, `chat.agentFilesLocations`, `chat.instructionsFilesLocations`, with `chat.useAgentSkills` defaulting to `true`), and the Copilot CLI agent runtime reads the same folders. No adapter wrote to any of them, so a VS Code user got nothing.

## What the adapter does

| Coco artifact | Target | Mechanism |
|---------------|--------|-----------|
| `skills/<name>/` | `~/.copilot/skills/<name>/` | symlink |
| `agents/*.md` | `~/.copilot/agents/*.md` | symlink |
| `commands/<ns>/<name>.md` | `<VS Code User>/prompts/<ns>-<name>.prompt.md` | symlink |
| `commands/<ns>/_index.md` | `<VS Code User>/prompts/<ns>.prompt.md` | symlink |
| `rules/cursor-mdc/*.mdc` | `~/.copilot/instructions/*.instructions.md` | generated |

One install serves both the editor and the terminal.

Three decisions worth flagging:

- **Commands are `<ns>-<name>`, not `<ns>:<name>`.** VS Code derives the slash-command name from the filename and a colon is not portable, so `/eng:anti-pattern` is `/eng-anti-pattern` here.
- **Prompt files go into every profile folder.** They are the one artifact with no user-level default outside the VS Code profile, so they are installed into `<User>/prompts` and each `<User>/profiles/*/prompts`, for VS Code, Insiders and VSCodium. Otherwise switching profile silently loses every command.
- **Rules are the only generated output.** VS Code needs the `.instructions.md` extension and an `applyTo` glob where the source is Cursor's `.mdc` with `globs` / `alwaysApply`. Everything else is a symlink and cannot go stale.

`--systems` is supported, including Super Intelligence — its SI-* commands are generated into `~/.copilot/coco-generated/si-commands` and linked from there.

Two adapter-specific flags: `--source`, for linking a checkout other than the one the script lives in (running from a git worktree that is later removed would otherwise leave every symlink dangling — this happened during development), and `--user-dir`, for an installation the detector cannot find.

## Bugs caught in review and fixed before merge

- `rules/cursor-mdc/react-default-skills.mdc` ships `globs: "**/*.jsx, **/*.tsx"` already quoted. Re-wrapping it produced `applyTo: ""**/*.jsx, **/*.tsx""` — invalid YAML, so VS Code dropped the rule silently. Values are now unquoted before being re-emitted, and escaped consistently.
- A CRLF checkout (Git for Windows default, and the adapter claims Windows support) made every `---` delimiter compare fail, so all frontmatter was dumped into the body and every rule became always-applied. The transform now strips trailing `\r`.
- The documented uninstall ran `rm -rf ~/.copilot/instructions`, a VS Code default folder that may hold the user's own instruction files. It now matches on the generated-by marker instead.

## Verification

- `bash tests/smoke.sh` — 19/19 pass, including the new adapter's dry run
- Installed for real on macOS against VS Code 1.134.0: 150 skills, 34 agents, 279 prompt files per profile, 15 instruction files, zero dangling symlinks
- All 15 generated instruction headers parse as YAML (they did not before the fix)
- `--dry-run` verified to write zero bytes; re-runs verified idempotent; no-VS-Code-installed path exits 0 and skips commands

Closes #4.


coco-research (approved):
Reviewed: symlink-only installer, no destructive operations. Hyphen namespace join is correct and documented (colon is not a portable filename). CI green.

PR #108, opened 2026-08-20
feat(si): make the local build endpoint configurable

rijul-mck:
## Why

`build_local.py` already speaks plain OpenAI chat-completions, but the endpoint was hardcoded:

```python
ENDPOINT = "http://127.0.0.1:1234/v1/chat/completions"   # LM Studio
```

That line is duplicated verbatim across **six** byte-identical copies (`finance`, `gtm`, `strategy`, `trading`, `risk-compliance`, `data-analytics`), so pointing the pipeline at any other local server meant editing all six.

## What changed

- `COCO_LOCAL_ENDPOINT` and `COCO_LOCAL_MODEL` are now read from the environment.
- Defaults are unchanged (`127.0.0.1:1234`, `qwen/qwen3.6-35b-a3b`), so existing runs behave exactly as before.
- An explicit `--model` still wins over `COCO_LOCAL_MODEL`.
- The resolved endpoint is printed next to the model — previously "which server did that actually hit" was invisible.

## Why it matters

This makes an [exo](https://github.com/exo-explore/exo) cluster a drop-in backend. exo serves the identical OpenAI shape on `:52415`, and unlike a single-process server it can span several machines, so the model is no longer capped by one box's RAM. `mlx-community/Qwen3.6-35B-A3B-8bit` is in exo's built-in catalogue and is the same base model the roster was built with.

```bash
COCO_LOCAL_ENDPOINT=http://localhost:52415/v1/chat/completions \
COCO_LOCAL_MODEL=mlx-community/Qwen3.6-35B-A3B-8bit \
  python3 finance/scripts/build_local.py --all
```

## Verification

- All 6 scripts compile clean under `-W error::SyntaxWarning` and remain byte-identical to each other.
- Precedence exercised end-to-end via the real CLI:
  - default → `qwen/qwen3.6-35b-a3b via http://127.0.0.1:1234/...`
  - env → `mlx-community/Qwen3.6-35B-A3B-8bit via http://localhost:52415/...`
  - `--model flag-model` beats `COCO_LOCAL_MODEL=env-model`
- `tests/smoke.sh` 18/18; `build-index.py` reports no INDEX drift.

One bug caught while reviewing my own diff: the usage example's trailing `\` was being consumed as a string line-continuation, silently collapsing the two env vars onto one line and yielding a broken copy-pasteable command. Escaped so it renders literally.

Note: `registry.json` / `roster.json` `build_method` fields still say "LM Studio" — those are historical records of how the roster was actually built, so they were deliberately left alone.

### DIFF
--- a/systems/superintelligence/README.md
+++ b/systems/superintelligence/README.md
@@ -71,6 +71,20 @@ Persona generation was an explicit cost/quality experiment, settled by evidence:
 3. **gemini-3.5-flash** (`build_cursor.py`, Cursor headless agent) — top quality, but agent-latency-flaky at batch scale.
 4. **Claude research agents (WebSearch)** — the quality winner; built the pending tail via a fan-out workflow.
 
+`build_local.py` talks plain OpenAI chat-completions, so it is not tied to LM Studio. Point
+it at any server speaking that shape — including an [exo](https://github.com/exo-explore/exo)
+cluster, which lets a model span several machines rather than being capped by one:
+
+```bash
+COCO_LOCAL_ENDPOINT=http://localhost:52415/v1/chat/completions \
+COCO_LOCAL_MODEL=mlx-community/Qwen3.6-35B-A3B-8bit \
+  python3 finance/scripts/build_local.py --all
+```
+
+Both variables are optional; unset, the defaults remain LM Studio and `qwen/qwen3.6-35b-a3b`,
+which is what the roster above was actually built with. An explicit `--model` still wins over
+`COCO_LOCAL_MODEL`. Whichever model you name must already be loaded by that server.
+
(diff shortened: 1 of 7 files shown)


rijul-mck:
Closing: this was written to support adopting exo, and we've decided against that.

exo's value is multi-device (RAM pooling, tensor parallelism, RDMA between peers). On a single machine every one of those is inert, and its built-in catalogue tops out at Qwen3.6 — there's no Qwen3.8 card — so adopting it would mean stepping back a model generation from what's already running in LM Studio. Not worth it.

The branch `feat/local-endpoint-config` is left in place, so if the hardcoded endpoint duplicated across the six `build_local.py` copies ever becomes a real annoyance, this can be reopened as-is. It changed nothing by default and CI was green.

PR #94, opened 2026-08-08
feat(agents): vendor 22 claude-code-templates agents with MIT attribution

rijul-mck:
## What this does

Vendors twenty-two third-party agent definitions from [davila7/claude-code-templates](https://github.com/davila7/claude-code-templates) into `agents/vendor/claude-code-templates/`, fetched from the upstream `main` branch via that project's CLI at version 1.29.4.

The agents cover backend and frontend development, architecture review, AI and prompt engineering, Rust, cloud and Terraform infrastructure, CI/CD and observability, security auditing and penetration testing, product management, and screenshot analysis.

## Why a nested vendor directory

Two reasons, both deliberate.

Four of the upstream filenames — `ai-engineer.md`, `code-reviewer.md`, `database-architect.md` and `ui-ux-designer.md` — collide with CoCo-authored agents of the same name that have been in `agents/` since April. Placing the vendored files flat would overwrite CoCo's own work. A nested directory avoids that entirely and additionally keeps the provenance and licence boundary explicit.

## Licence compliance

Upstream is MIT, Copyright (c) 2025 Daniel (San) Ávila. MIT permits redistribution only when the copyright notice and permission notice accompany the files, so this PR:

- retains the upstream notice verbatim at `agents/vendor/claude-code-templates/LICENSE`;
- adds a `CREDITS.md` row naming the project, author, scope, upstream URL and licence, matching the format of the existing entries;
- documents full provenance in the vendor directory's `README.md`, including the upstream branch, CLI version, fetch date, the upstream category of every agent, and the re-sync procedure.

CoCo's root MIT licence covers CoCo-authored material and does not extend to this directory.

## Two known quirks, both upstream's

`architect-review.md` declares `name: architect-reviewer` in its front matter. Claude Code identifies an agent by the `name` field rather than the filename, so it must be invoked as `architect-reviewer`. Left as-is to keep the vendored files byte-identical to source.

These agents do not appear in `agents/INDEX.md`. The generator at `scripts/build-index.py` globs `agents/*.md` and `systems/*/agents/*.md`, neither of which matches a nested directory. The generator was deliberately left unmodified so that vendoring third-party content does not change how CoCo's own agent catalogue is built; the vendor `README.md` is the catalogue for these twenty-two instead.

## Verification

- All 22 files present with valid YAML front matter.
- CoCo's four same-named authored agents in `agents/` confirmed unmodified (checksums match pre-change values).
[...]

PR #109, opened 2026-08-20
feat(skills): vendor the reverse-skill bundle, defensive subset only

rijul-mck:
## Summary

- Vendors a curated, reference-only subset of [zhaoxuya520/reverse-skill](https://github.com/zhaoxuya520/reverse-skill) at `systems/reverse-skill/` — thirty-six reverse-engineering and security methodology skills, fetched from upstream `main` at commit `a3bdfff`.
- Adds `skills/coco-ship/`, a new skill formalizing the branch → commit → guard-aware push → PR → CI-wait flow used by hand for [#94](https://github.com/coco-research/coco/pull/94), so shipping to a coco-managed repo is repeatable rather than re-derived each time.

## What the audit found, and what's excluded

Upstream is mostly legitimate: solid methodology, an honest self-audit with a real hardening changelog (`docs/PACKAGE-SECURITY-AUDIT.md`), and its own `skills/ops/skill-supply-chain.md` naming the exact risks a careful third-party skill vendoring should check for.

It also has a router/compliance layer — `RULES.md`, `CLAUDE.md`, `AGENTS.md`, the top-level skill router — explicitly written to make a coding agent skip confirmation and act immediately ("if you only reply 'understood'... you have failed... do not wait for the user to say ok continue"). `skills/llm-security/references/agent-obedience-engineering.md` is a dedicated methodology doc for exploiting exactly that behavior in coding agents. Separately, `skills/ida-reverse/scripts/install-autostart.ps1` registers a hidden Windows scheduled task that respawns a watchdog every minute for 3650 days.

None of that is vendored. For every included topic, only `SKILL.md` and `references/**/*.md` were copied — no `scripts/`, no agent persona configs. It turned out the same compliance-ritual template (an immediate-action block at the top, a router-context and self-check block at the bottom, a bootstrap block with real install commands on some files) is stamped onto every single upstream `SKILL.md`, woven into the same document as the real content — that was stripped from every file kept here, not just excluded from copying. `systems/reverse-skill/README.md` has the full provenance and inclusion/exclusion rationale, including a short list of incidental (non-executable, non-directive) dangling references left in a few otherwise-substantive reference docs.

License is genuine MIT (GitHub's API just failed to detect it) — notice retained at `systems/reverse-skill/LICENSE`, `CREDITS.md` updated to match.

## coco-ship

Direct formalization of Option 2 ("push and create PR") in the existing `skills/finishing-a-development-branch/SKILL.md`, scoped to repos carrying the corporate git-guard:

- Detects whether a `GIT_GUARD_REASON` bypass is actually needed (checks repo visibility/owner and whether `core.hooksPath` even points at the guard) rather than assuming.
- Asks for explicit confirmation — repo, branch, guard reason, and that this opens a PR rather than pushing to `main` — every time, regardless of how it's invoked.
- Waits for real CI before reporting done.
- Never merges, never self-approves, never pushes to `main` directly.

This PR was itself shipped by following that process by hand (the skill only registers with the harness at session start, so this run predates its own first live invocation).

## Test plan

- [x] `find systems/reverse-skill -type f` restricted to `*.md` only — confirmed, zero scripts/binaries.
- [x] Zero `scripts/` or `agents/*.yaml` subdirectories present anywhere in the bundle.
- [x] Compliance-ritual template (`ACTION REQUIRED`, `路由上下文`/`路由注册`, `按需自举`/`On-Demand Bootstrap`, `任务完成自检`) stripped from all 36 topic `SKILL.md` files — verified with a level-and-fence-aware script, not a naive heading match.
- [x] `agent-obedience-engineering.md` absent; the one `SKILL.md` line that named and described it removed too.
- [x] McKinsey-internal, Atlassian, and secret-shaped-string scan across the full bundle: clean.
- [x] `scripts/build-index.py` re-run; `skills/INDEX.md` lists all 36 new topics under "Bundle: reverse-skill" plus `coco-ship` as a core skill.
- [x] `agents/vendor/claude-code-templates/` (from #94, still open) confirmed absent from this branch — the two PRs are independent.


coco-research (approved):
Reviewed and verified locally: both CI workflows pass on this head (ci.yml 11/11 steps, lint-frontmatter 3/3). #94 is rebased onto current main with no INDEX drift; #109's conflict was confined to the generated skills/INDEX.md and is resolved by regenerating rather than by hand. Branch is mergeable with no failing checks.

PR #111, opened 2026-08-20
docs(local-llm): task-fit, failure modes, web access, and a tested client

rijul-mck:
## Summary

Four gaps found by *using* the `local-llm` skill in anger rather than reading it. Every new section was **drafted by the local model itself** and then verified against measured output.

- **What this model is good at, and what it is not** — the doc explained how to call the endpoint but never what to send it. Records strong uses (bulk diff summarisation, classification, mechanical transforms) and weak ones (merge, security, licensing calls).
- **Known failure modes** — four *reproduced* failures with mitigations:
  - **Diff polarity inversion.** The model reported a change in the wrong direction on a real diff. Had it been trusted, the conclusion would have been backwards. Mitigation: state polarity twice, require a verbatim quoted diff line as evidence per claim.
  - **Silent truncation** (`completion_tokens == max_tokens`) — looks finished, isn't.
  - **Ignored reasoning suppression** — the original hang.
  - **Stop-token collision.** Asking the model to echo a code block containing its own `<|im_end|>` halted generation at 112 tokens, indistinguishable from ordinary truncation. Hit while writing this PR.
- **Giving the model web access** — states plainly that the server has no network egress, documents the orchestrator-fetch pattern that works today vs. a tool-calling loop that isn't worth building yet.
- **Calling it from your own script** — `build_local.py` is a persona pipeline, not a reusable client, so every caller had to reconstruct the prefill request from prose. Adds a standalone client plus measured timings (prompt processing dominates; a 13.8k-token prompt took ~2 min).

## Test plan

- [x] `scripts/build-index.py` clean (185 skills / 38 commands)
- [x] `tests/check-command-refs.sh` PASS
- [x] `tests/smoke.sh` — **19 passed, 0 failed**
- [x] Every `python` block in the file compiles
- [x] **The documented client was extracted from the file and executed verbatim against the live server**, returning the expected completion. The code in the doc is the code that ran.
- [x] Zero em dashes (house style)

### DIFF
--- a/skills/local-llm/SKILL.md
+++ b/skills/local-llm/SKILL.md
@@ -156,6 +156,96 @@ Measured against that ceiling:
 5. Hardware upgrade (M4 Ultra ~= 2x memory bandwidth) -- out of scope, just
    noted for completeness.
 
+## What this model is good at, and what it is not
+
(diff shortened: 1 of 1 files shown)


coco-research (approved):
Documented client executed verbatim from the file against the live server; all gates pass, smoke 19/0.

PR #116, opened 2026-08-20
fix(local-llm): measured throughput figures, and how to use the model efficiently

rijul-mck:
## Summary

The speed numbers in this skill were wrong **and internally inconsistent**: it claimed 20–35 tok/s while also quoting 750 tokens in 133 s, which is a 5.6 tok/s ceiling before prefill is even counted. Both could not be true. The underlying error was mine, reporting wall clock as though it were decode speed.

Remeasured single-stream, with the `/metrics` request counter checked before and after every call so nothing else was competing.

### Decode is not one number

| Output kind | Accepted per round | Time per round | Rate |
|---|---|---|---|
| Trivially predictable (counting) | 8.00 | 78 ms | **103 tok/s** |
| Open-ended prose | 2.55 | 133 ms | **19.3 tok/s** |

The mechanism is now **measured, not assumed**. The server drafts up to 8 tokens per verification round and pays one full pass per round regardless of how many survive. Fewer accepted tokens *and* slower rounds compound, and the arithmetic closes on the observed spread. Server-reported mean across 30 mixed real requests is ~14 tok/s, which is the figure to plan against.

Prefill is a separate cost of ~**183 tok/s**, derived from two requests differing by 5,995 prompt tokens (32.7 s apart). A 13k-token prompt therefore costs ~70 s before generation starts.

Also corrects the older speed cheatsheet, which asserted the same wrong range.

### Concurrency actively hurts

| Concurrent | Aggregate tok/s | Mean per request | Slowest |
|---|---|---|---|
| 1 | **12.7** | 12.7 | 12.7 |
| 2 | 12.4 | 9.6 | 6.1 |
| 4 | **11.3** | 6.1 | **2.2** |

Aggregate throughput *falls* as callers are added, which is backwards from normal batching. Fairness collapses too. Documented with the operational conclusion (serialise, don't fan out) and a memory note: each batch slot reserves ~4 GB of KV cache at this context length, so lowering `--max-batch` would free ~12 GB. That file lives outside the repo and **has not been changed**.

### Client hardened

Replaced with a version **executed from this file against the live server**. It now fails loudly rather than silently when a reply is cut short, covering both the case where the server omits `usage` (previously `None == max_tokens` was false, so truncation passed silently) and a stop-string collision that halts well below `max_tokens`. Accepts `stop=()` for output that may legitimately contain the stop string.

## Test plan

- [x] Every figure measured on this machine, single-stream, request counter verified
- [x] Documented client extracted from the file and run: correct reply, and truncation guard raises
- [x] `build-index.py` 185/38/34 · `check-command-refs.sh` PASS · `smoke.sh` **19/0**
- [x] Zero em dashes

### DIFF
--- a/skills/local-llm/SKILL.md
+++ b/skills/local-llm/SKILL.md
@@ -137,10 +137,15 @@ Measured against that ceiling:
 - LM Studio: 25-27.5 tok/s (74-81% of ceiling -- well-optimized).
 - mlx-dspark baseline (non-speculative): 15.2-19 tok/s (45-56% -- less
   optimized kernels than LM Studio's llama.cpp-derived backend).
-- mlx-dspark `dflash` (speculative decoding): 20-35.9 tok/s (59-106% --
-  *can exceed* the naive per-token ceiling, because speculative decoding
-  amortizes one expensive full-weight read over `accept_len` ~2.7 accepted
-  tokens per verification round instead of one token per read).
+- mlx-dspark `dflash` (speculative decoding): highly variable, because the
+  gain depends entirely on how predictable the output is. Measured on this
+  machine at 111 tok/s on trivially predictable output and 10.1 tok/s on
+  open-ended prose, against a server-reported mean of about 14 tok/s across
+  mixed real requests. Speculative decoding amortizes one expensive
+  full-weight read over several accepted draft tokens, so it can exceed the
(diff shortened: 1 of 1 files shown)


coco-research (approved):
Every figure measured on this machine; documented client executed verbatim from the file after each change; truncation and 500/503 recovery both exercised.

PR #118, opened 2026-08-22
feat: add Hermes Agent adapter

coco-research:
Adds a first-class `adapters/hermes/` adapter that wires Coco artifacts into every Hermes Agent profile.

## What
- `skills/<name>/` → `<profile>/skills/<name>/` (symlink — Hermes discovers skills here)
- `systems/<bundle>/skills/*` → `<profile>/skills/*` (via `--systems`)
- `agents/*.md` → `<profile>/home/.claude/agents/*.md`
- `commands/*.md` → `<profile>/home/.claude/commands/*.md`
- rules → idempotent marker-wrapped block appended to `<profile>/home/.claude/CLAUDE.md`
- Super Intelligence `/SI-*` commands generated into `<profile>/home/.claude/commands`

## Usage
```bash
# one profile
bash adapters/hermes/install.sh --profile dev
# every profile on the machine
bash adapters/hermes/install.sh --all-profiles --systems gsd,brain,team,superintelligence,cognee,hyperframes,m0
```

## Verification
- `install.sh --dry-run` works
- Full activation: 171 skills / 24 agents / 242 `/SI-*` commands — matches the claude-code adapter exactly
- CI gates pass: manifest validation, bundle-install-plan gate (`--systems` provably changes the plan)
- Idempotent: re-run reports zero STALE conflicts

## Docs
- Added `hermes` to the adapter table in `docs/architecture.md`

PR #119, opened 2026-08-22
fix(ci): cover hermes/vscode adapters in lint and smoke gates

coco-research:
## What
- Extend KNOWN_ADAPTERS in lint-frontmatter.yml: claude-code, cursor, codex, generic → + hermes, vscode. Previously any SKILL.md declaring supports:[hermes] would fail lint (verified by simulation; currently dormant because no skill declares them yet).
- Add hermes to the adapter dry-run smoke loop in ci.yml AND tests/smoke.sh (third inline copy found per review).
- Both loops guarded with [ -d adapters/ ] || SKIP so this lands cleanly on main before #118 merges. Guards are self-removing once #118 is on main.

## Verification
- tests/smoke.sh on main-state tree: 19 passed / 0 failed, hermes SKIPs cleanly.
- Same suite on a tree with adapters/hermes present: 20/20 PASS.
- bash -n syntax checks pass.

## Review
Approved with amendments by @dev (A1: patch all inline copies; ordering decision: exists-guard over merge-order inversion). Follow-up task to strip guards recorded.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,9 @@ jobs:
 
       - name: Adapter smoke test (dry-run)
         run: |
-          for adapter in claude-code cursor vscode codex generic; do
+          for adapter in claude-code cursor vscode codex generic hermes; do
+            # hermes ships via PR #118; skip until it lands on main.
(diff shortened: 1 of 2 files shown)


rijul-mck (approved):

PR #117, opened 2026-08-21
fix(security): close eval, rm-rf, uninstall-glob, and unpinned-clone holes

coco-research:
## Summary

Closes the install/skill/brainstorm holes found in the security review. One PR, six surfaces. Please **do not auto-merge** — leave merge to the brain agent.

1. **`generate_prd.sh` no longer `eval`s user input.** Interactive answers used `eval "$var_name='$input'"`, so a value containing a quote ran as shell. Assignment is now `printf -v` into an allow-listed variable name. The regression check feeds `'; echo PWNED; x='` and asserts it is stored as a string.
2. **`stop-server.sh` canonicalizes before the `/tmp` prefix check.** `[[ $path == /tmp/* ]]` is a glob, so `/tmp/../Users/you` matched and `rm -rf` followed. The path is resolved with `pwd -P` and compared against the real `/tmp` root (`/private/tmp` on macOS). Only a session directory under that root is deleted. PIDs must be numeric before `kill`.
3. **Uninstall matches the clone path plus a trailing slash.** `find -lname '*${dir}*'` deleted links into `coco-research` / `coco-connect` when the clone was `coco`. `bin/coco.js` and the docs now match `${clone}/*`. The regression check plants both a `coco` and a `coco-research` symlink and asserts only the former is removed.
4. **Skills no longer instruct `curl | sh` or `npx skills add -y`.** `browser-automation` and `ai-marketing-videos` fetch the inference.sh installer to a file for review. `find-skills` installs without `-y` and still requires an explicit yes that names the package.
5. **Fresh CLI clones are pinned to the release tag, not floating `main`.** `bin/coco.js` and `bin/coco-bootstrap.sh` clone `v1.2.0` (kept in sync with `package.json`). Existing *branch* checkouts still `git pull --ff-only`, so a tagged pin cannot downgrade a `main` tree. Detached tagged installs re-fetch the pin.
6. **Documented that `cocosuperintelligence` is unpublished.** The name 404s on npmjs.com. README, `coco-cli`, and `SECURITY.md` tell users not to `npx` it until it is published — npx would execute whatever first claims the name.

`tests/check-security-surface.sh` covers (1)–(5) and is wired into `tests/smoke.sh` (already run by CI).

Not in this PR (still open, separately): Outlook/LaunchAgent scan in `commands/pm/sync-init.md`; unauthenticated brainstorm `server.js` bind-to-0.0.0.0; Homebrew formula still pins the v1.2.0 tarball (correct for a formula — bump on the next release).

[...]

PR #120, opened 2026-08-22
docs: complete adapter table, fix CoC reporting line (#79), expand uninstall FAQ

### LINKED ISSUE #79: CODE_OF_CONDUCT.md points reporters at a `private` label that does not exist
`CODE_OF_CONDUCT.md` line 19 says:

> Concerns can be reported by [opening an issue](https://github.com/coco-research/coco/issues) (use the `private` label or email if sensitive).

**The repo has no `private` label.** Verified via `gh api repos/coco-research/coco/labels`, the full set is: `bug`, `documentation`, `duplicate`, `enhancement`, `good first issue`, `help wanted`, `invalid`, `question`, `wontfix`. So the documented reporting path cannot be followed as written.

There is a second, more important problem with the instruction itself: **a label cannot make an issue private.** On a public repository, an issue filed with a `private` label is still world-readable, so following this instruction would cause a reporter to disclose a sensitive Code-of-Conduct concern publicly — the opposite of what the sentence intends.

Suggested fix (rewording rather than adding the label):

- Remove the `private`-label suggestion entirely.
- Point sensitive reports at a genuinely private channel: a maintainer email address, or GitHub's **private vulnerability reporting** / a security advisory, which is private by design.
- Keep public issues as the path for non-sensitive conduct questions only.

Adding a `private` label would make the doc literally accurate but would preserve the misleading implication that the label confers privacy, so rewording is the safer resolution. Flagging rather than deciding, since it changes a published policy document.

coco-research:
## What

This pull request originally predated a security fix that landed on main after it was opened, and its own uninstall content would have reintroduced that defect. It has been merged forward onto the current main and repaired.

- docs/architecture.md: the adapter table now matches the `adapters/` directory exactly. It already carried this pull request's `hermes` and `vscode` rows plus main's own `grok` row from a parallel pull request; this update adds the eight adapters that existed in the tree but were still missing from the table, `aider`, `amazon-q`, `cline`, `github-copilot-cli`, `roo-code`, `vscode-continue`, `windsurf`, and `zed`.
- CODE_OF_CONDUCT.md: the reference to a nonexistent `private` label is removed (closes #79); the reporting line now points to opening an issue or emailing a maintainer directly, with no personal name or corporate email address anywhere in the file. Confirmed with `git grep -ci rijul` (zero matches) and `git grep -in mckinsey` (55 matches, all persona biography content unrelated to the reporting line).
- README.md uninstall FAQ: this is the security-relevant change. Main had already replaced this pull request's unanchored `find ... -lname "*$(pwd)*" -delete` contains-match with an anchored, clone-path-prefixed form (`CLONE="$(pwd)"; find ... -lname "${CLONE}/*" -delete`) across PR #117 and PR #178, because a contains-match deletes symlinks into any sibling directory whose path merely contains the clone path as a substring, for example a `coco-research` clone next to a `coco` clone. This merge keeps main's fixed primary find line unchanged and adds this pull request's genuinely new Hermes-specific uninstall paragraph using the same anchored shape, with its own `CLONE="$(pwd)"` assignment repeated inside that same `<pre>` block rather than reused from the block above, so the snippet is copy-paste safe on its own.

## Verification

Every uninstall block in the final tree (`README.md` x2, `docs/TROUBLESHOOTING.md`, `docs/install.md`, `adapters/claude-code/README.md`, `adapters/hermes/README.md`, `adapters/vscode/README.md`, `adapters/grok/README.md`) was checked for two properties: the `-lname` glob anchors the clone path as a prefix with a trailing separator, and every variable it uses is assigned inside that same block. All pass.

Each block was then extracted verbatim and run in isolation, in a throwaway directory with `HOME` and the working directory pointed at a fresh adversarial layout containing a `coco` clone and a sibling `coco-research` clone, each with a symlink into it from every relevant target directory. Every block deleted only the link into the `coco` clone and left the `coco-research` link untouched, and none expanded to a bare `/*`. Example, the README's Hermes-specific block run alone: before, `.hermes/profiles/coco-link` and `.hermes/profiles/research-link` both exist; after running only `CLONE="$(pwd)"; find ~/.hermes/profiles -type l -lname "${CLONE}/*" -delete`, only `research-link` remains, and links under `.claude`, `.cursor`, `.grok`, and `.copilot` from the same adversarial layout are untouched because the block never referenced them.

The generated files (`skills/INDEX.md`, `commands/INDEX.md`, `agents/INDEX.md`, `systems/INDEX.md`, `docs/by-domain/*.md`, `docs/asset-counts.json`) were regenerated with `python3 scripts/build-index.py` rather than hand-merged, and a second run produced an identical result.

All six required gates pass on the merged tree: `python3 scripts/build-index.py`, `bash tests/check-command-refs.sh`, `bash tests/check-evidence-gate.sh`, `bash tests/check-asset-counts.sh`, `bash tests/check-security-surface.sh`, and `bash tests/smoke.sh`. `tests/check-security-surface.sh` is the decisive gate for this pull request; its class-wide check for any unanchored contains-match link glob anywhere in the tree passes cleanly.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -16,7 +16,7 @@ Anyone whose actions consistently make the project less welcoming or less produc
 
 ## Reporting
 
-Concerns can be reported by [opening an issue](https://github.com/coco-research/coco/issues) (use the `private` label or email if sensitive). Reports are handled discreetly.
+Concerns can be reported by [opening an issue](https://github.com/coco-research/coco/issues) or emailing a maintainer directly if sensitive. Reports are handled discreetly.
 
 ## External reference
 
--- a/README.md
+++ b/README.md
@@ -718,8 +718,13 @@ Every skill lives at <code>skills/&lt;name&gt;/SKILL.md</code>. Edit it directly
 <details>
 <summary><strong>How do I cleanly uninstall CoCo?</strong></summary>
 Because CoCo uses symbolic links, removal is non-destructive:
-<pre>find ~/.claude ~/.cursor -type l -lname "*$(pwd)*" -delete</pre>
+<pre>find ~/.claude ~/.cursor ~/.copilot -type l -lname "*$(pwd)*" -delete</pre>
 This removes only the links pointing back to your CoCo repository folder.
+
+For the Hermes adapter, profiles live under <code>~/.hermes/profiles</code>:
+<pre>find ~/.hermes/profiles -type l -lname "*$(pwd)*" -delete</pre>
+The appended rules block between <code>&lt;!-- coco:rules-start --&gt;</code> and
+<code>&lt;!-- coco:rules-end --&gt;</code> in each profile's CLAUDE.md is stripped automatically on the next install run without the repo.
 </details>
 
 <details>
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -103,11 +103,13 @@ Current first-class adapters (v1.0.0):
 | Slug | Target | Status |
 |------|--------|--------|
 | `claude-code` | Anthropic Claude Code CLI | stable |
+| `hermes` | Hermes Agent profiles (all bots) | stable |
 | `cursor` | Cursor IDE | stable |
 | `codex` | OpenAI Codex CLI | stable |
 | `generic` | Any tool reading `AGENTS.md` | stable |
+| `vscode` | VS Code / Copilot CLI (`~/.copilot`) | stable |
 
-`supports: [claude-code, cursor, codex, generic]` declares an artifact works everywhere. Omit an adapter to mark unsupported.
+`supports: [claude-code, cursor, codex, generic, hermes, vscode]` declares an artifact works everywhere. Omit an adapter to mark unsupported.
 
 ## Skill folder structure
 


rijul-mck (approved):

rijul-mck (approved):
Verified. The defect this pull request originally carried is resolved the right way round: where its README side and main disagreed about an uninstall command, main's anchored form was kept unchanged and the new Hermes paragraph was re-added on top using the safe shape, rather than the pull request's unanchored pattern being merged back in.

The property that the earlier proposed fix got wrong is now explicitly satisfied. Each uninstall block assigns its own clone variable inside the same block that uses it, so a reader copying any single block gets a working command rather than one that expands to a bare wildcard. That was proven per block by running each one in isolation against a layout containing both a coco clone and a sibling coco-research clone, and confirming only the coco links were removed.

The adapter table was missing 8 of the 15 real adapters and now lists exactly the directories that exist, with descriptions taken from each adapter's own manifest rather than invented.

On the code of conduct change, the reporting line drops a label that does not exist and uses role-based phrasing that discloses no address at all, which is the right outcome. I confirmed the owner's personal name appears nowhere in the tree. The 55 matches for the employer name are all third-party persona biographies, which are legitimate and must be left alone.

Noting for the record that tests/check-security-surface.sh currently produces a false positive in any checkout that has nested worktrees under .claude/worktrees/ or an untracked file quoting the unsafe pattern, because it scans the working tree rather than tracked files. That is a defect in the gate rather than in this change, and it is being fixed separately.

PR #122, opened 2026-08-22
feat(ci): asset-count single source of truth (docs/asset-counts.json + drift gate)

coco-research:
## Problem
Shipped surfaces disagreed on skill/command counts (README badge 149 vs hero 185 vs package.json 179/279 vs spec table 149). `docs/asset-counts.json` was missing.

## Merge bar (Mira, 2026-08-29)
Customer-facing commands = 38 shipped files + 242 `/SI-*` generated at install = **280**. Do not invent a third number. Do not stamp public commands from a walk of `commands/*.md` (that is 38, not the public figure).
Skills customer-facing = **185** (70 core + 115 bundle). Do not leave 149 as a third skills number.

SSoT labeled fields in `docs/asset-counts.json`:
- `skills.total` = 185 (core 70 + bundle 115)
- `commands.shipped` = 38 (`commands/*/*.md`)
- `commands.generated_si` = 242 (generated at install, not in repo)
- `commands.customer_facing` = 280 (sum; package.json, `.claude-plugin.json`, and README badges/spec)

## Fix
- Generator emits the two command layers from the tree walk + SI registry/generators.
- CI gate fails if public surfaces stamp 38 as the command total, or if README spec Total Skills (or leftover 149) disagrees with skills.total.
- package.json / plugin / README badges / spec table: **185 skills, 280 commands**. README 280 stays.

## Out of scope
GitHub About stays with Nia until Ryan's GitHub sign-in. Website copy is Nia's. Ryan is the review; coco-research cannot self-approve.

PR #124, opened 2026-08-30
feat(cursor): --systems superintelligence writes SI commands to ~/.cursor/commands

coco-research:
## Problem
`adapters/cursor/install.sh` accepted only `--dry-run`/`--help`. Any other flag was `Unknown flag` / exit 1. Auto-detect picks `cursor` when `~/.cursor` exists, so the README flagship command `bash install.sh --systems superintelligence` died before any adapter that supports `--systems` ran. Cursor users were generating SI files by hand (or being sent through the claude-code adapter).

## SUCCEEDS-WHEN (Annu)
- `bash install.sh --adapter cursor --systems superintelligence` exits 0 and writes SI command files (including `SI-Decide.md`) into `~/.cursor/commands` (overridable via `CURSOR_HOME`)
- `bash install.sh --systems superintelligence` must not fail solely because `~/.cursor` exists
- When both `~/.cursor` and `~/.claude` exist, `--adapter cursor --systems superintelligence` still targets Cursor

## Fix
- Cursor adapter parses `--systems` and generates SI-* with `COCO_SI_COMMANDS_DIR=$CURSOR_HOME/commands` (same generators as claude-code, different target)
- `supports_systems` lists the same bundles as the other adapters so the CI bundle gate is honest
- `tests/cursor-si-commands.sh` (run from smoke) asserts `SI-Decide.md` is written under `CURSOR_HOME`

Not folded into #122. Site copy untouched. Local-first; no telemetry.

## Test plan
- [ ] CI `lint-and-test` (bundle gate + smoke cursor SI step)
- [ ] `CURSOR_HOME=/tmp/c bash adapters/cursor/install.sh --systems superintelligence` → `/tmp/c/commands/SI-Decide.md`
- [ ] `bash install.sh --adapter cursor --systems superintelligence` still writes Cursor, not `~/.claude`

### DIFF
--- a/adapters/cursor/install.sh
+++ b/adapters/cursor/install.sh
@@ -4,16 +4,20 @@
 # Usage:
 #   bash adapters/cursor/install.sh
 #   bash adapters/cursor/install.sh --dry-run
+#   bash adapters/cursor/install.sh --systems superintelligence
+#   bash adapters/cursor/install.sh --systems gsd,brain,superintelligence
 
 set -euo pipefail
 
 REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
 TARGET_HOME="${CURSOR_HOME:-$HOME/.cursor}"
 DRY_RUN=0
+SYSTEMS=()
 
 while [[ $# -gt 0 ]]; do
   case "$1" in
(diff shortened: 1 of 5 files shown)


rijul-mck (approved):

PR #138, opened 2026-09-04
feat(si): wire 4 new departments into the registry and quality pipeline

coco-research:
## Summary

This pull request wires the four already-merged Super Intelligence departments (Legal & IP, Healthcare & Life Sciences, Education & EdTech, and Climate & Energy) into the shared registry and quality tooling. Each department's personas already existed on `main` from earlier, separately merged pull requests; what was missing, and what this pull request adds, is the central `systems/superintelligence/registry.json` entry for each department, the corresponding `build_commands.py` team entries, per-department `registry.json`/`cells/`/`SKILL.md`/`build_registry.py` infrastructure, and new quality-gate tooling (`quality_score.py`, `llm_judge.py`, `validate_offline.py`, a shared quality rubric template) together with refreshed `ROSTER.md` and `VALIDATION` files for the nine pre-existing teams.

## What changed since the original submission

An automated review found that this pull request, as originally submitted, could not be merged and needed repair before it was safe to publish. Two problems were confirmed independently against the tree, and both are fixed in the latest commit.

**Leaked personal path, fixed by deleting the generated directory.** The branch had committed `commands/si-generated/`, a directory of 342 files, of which 303 embedded another contributor's personal absolute checkout path 1,294 times. This was a regression of the exact defect class that pull request #86 already fixed once. Before deleting anything, the directory's origin was verified: `systems/superintelligence/ai/scripts/build_commands.py` is the generator (every file in the directory carries the header "Generated by `build_commands.py` — do not hand-edit; edit the generator and regenerate"), and the generator's default output location is `~/.claude/commands` (or `$COCO_SI_COMMANDS_DIR`), not a path inside the repository; it deliberately derives the repository path from its own file location at generation time (`COCO_SI_REPO` override, or `Path(__file__).resolve().parents[3]`) precisely so that the generated command bodies point at whichever machine installs them, rather than at one contributor's fixed checkout. The install adapters (`adapters/claude-code/install.sh` and the equivalent for Cursor, VS Code, and Hermes) already call this generator at install time. A repository-wide search for `commands/si-generated` across `scripts/`, `tests/`, `.github/`, `install.sh`, every adapter, `README.md`, and `package.json` found zero references outside the directory itself, confirming nothing depends on the committed copy. The directory has been deleted in full.

Deleting it changes the derived counts in `docs/asset-counts.json` in an expected way, not an unexpected one: this pull request's actual purpose is marking four more teams `built: true` in the registry, and `commands.generated_si` is computed live from the count of built teams, so it correctly rises from 242 to 342 (and `commands.customer_facing` from 286 to 386) once those four teams are wired in. `commands.shipped` (44) and `commands.namespaces` (7) are unaffected, matching the pre-existing baseline. `README.md`, `package.json`, and `.claude-plugin.json` are updated to match that same derived total so that `tests/check-asset-counts.sh` passes; no persona or department total is hand-stamped anywhere in this change.

**Count restamps removed; counts are left to the derived source of truth.** The original submission's title and description stamped a persona total ("495 personas total", "495/495 PASS, avg 96.4"). A separate, still-in-flight change is making `docs/asset-counts.json` the single derived source of truth for persona, department, and adapter counts, gated against every published claim. A hand-stamped persona total in this pull request would conflict with that effort and reintroduce the exact kind of drift it exists to prevent. This pull request now stamps no persona or department total anywhere; the title and this description say so explicitly, and the only counts touched are the pre-existing, already-gated skills/commands totals described above.

## Verification

After merging current `main` (which had moved significantly since this branch diverged, including all four department directories landing independently) and resolving the merge by regenerating every generated file with `python3 scripts/build-index.py` rather than hand-merging, the leak was confirmed absent by every one of the following, run against the final tree:

```
$ git diff origin/main...HEAD | grep -c '/Users/'
0

$ grep -rn --exclude-dir=.git -e '/Users/rijulkalra' -e '/Users/Rijul_Kalra' . | grep -v node_modules
(no output)

$ grep -rc --exclude-dir=.git '/Users/' . | grep -v ':0$' | grep -v node_modules
(only pre-existing, unrelated generic example paths inherited from main, e.g. /Users/jesse/... in
skills/using-git-worktrees/SKILL.md and skills/systematic-debugging/*, /Users/runner in a GitHub Actions
comment, /Users/<user>/ and /Users/xxx/ placeholders in GSD skill docs — none reference this contributor
or the leaked checkout path, and none were touched by this pull request)

$ git grep -n -e '/Users/' -- .
(same generic, pre-existing, unrelated matches as above; zero occurrences of the leaked path)

$ git grep -ci rijul -- .
(no output, exit 1 — zero matches anywhere in tracked content)
```

All six required gates pass on the final tree:

- `python3 scripts/build-index.py` — regenerates cleanly, idempotent (a second run produces no further diff)
- `bash tests/check-command-refs.sh` — PASS: all command cross-references resolve
- `bash tests/check-evidence-gate.sh` — PASS: `/team` evidence-gate language present
- `bash tests/check-asset-counts.sh` — PASS on every check (README badges, package.json, README prose table, README spec table, `.claude-plugin.json`, all matching the regenerated `docs/asset-counts.json`)
- `bash tests/check-security-surface.sh` — PASS on every check
- `bash tests/smoke.sh` — 23 passed, 0 failed

The owner's personal name does not appear anywhere in the tracked tree (confirmed above); it was previously removed from all 17 places it had appeared, and this pull request keeps it absent.

## Note on counts

This pull request deliberately does not stamp any persona or department total. Those numbers are left to the derived source of truth (`docs/asset-counts.json`) that a separate, still-in-flight change is extending to cover personas, departments, and adapters. Only the pre-existing, already-gated skills/commands totals are updated here, to keep `tests/check-asset-counts.sh` passing after this pull request's own registry integration legitimately changes how many teams are built.


rijul-mck (approved):

rijul-mck (approved):
Verified independently on the pushed head.

The leak is gone. commands/si-generated/ is absent, and it held 342 files of which 303 carried an absolute personal path, 1294 occurrences in total. Deletion was the right remedy rather than path stripping, because the generator at systems/superintelligence/ai/scripts/build_commands.py derives the repository root from its own location at generation time and is already invoked by the adapters at install, so the committed copy was redundant output. A tree-wide search found nothing depending on it. The diff carries no absolute user paths and the owner's name appears nowhere in tracked content or in any commit message unique to the branch.

The skills total moving from 222 to 226 is explained and derived rather than stamped. Four new team SKILL.md files land for climate-energy, education-edtech, healthcare-life-sciences and legal-ip, none of which exist on main, and build-index.py counts them through its systems/*/*/SKILL.md glob. The commands total moving to 386 follows from the same four teams becoming built.

I checked the site pages myself, since no gate covers them yet. Neither page retains any stale figure, and the persona figure of 389 is preserved exactly: it appears five times on the removed lines and five times on the added lines, with only the adjacent skills number changed. That leaves the persona and department corrections cleanly to the change that owns them.

Three persona files with genuinely malformed YAML frontmatter are repaired here. I confirmed all three fail to parse on main, so this is a real defect being closed rather than a preference. Worth recording that they reached main because the frontmatter lint targets a self-hosted runner that is not picking up jobs, so that check queues indefinitely and has never actually gated anything.

Finally, a completed local work-tracking file has been dropped from the change and added to .gitignore, which keeps private working state out of a public repository.

PR #141, opened 2026-09-06
fix(brain): replace inline code generation with stdin-based chroma_worker

coco-research:
## Summary
- Adds `chroma_worker.py`: subprocess worker that reads JSON from stdin for ChromaDB operations (upsert, get_ids, stats)
- Updates `memory_bridge.py`: all 4 f-string script generation sites replaced with subprocess calls piping JSON to chroma_worker
- Eliminates arbitrary code execution risk from dynamically generated Python strings

## Security Impact
Previously, memory_bridge.py constructed Python code as f-strings and executed them via subprocess. This created an injection vector if any input contained unexpected characters. The new approach uses structured JSON over stdin, making injection impossible.


rijul-mck (approved):
Reviewed in batch: CI green, mergeable, scanned for owner name, employer and client names, licence-restricted paths, NC licences, credential shapes and self-hosted runner changes.

PR #142, opened 2026-09-06
fix(installer): add target prefix validation to prevent symlink traversal

coco-research:
## Summary
- Adds case check in `link_dir()` within `adapters/claude-code/install.sh` to refuse creating symlinks outside $TARGET_HOME
- Prevents path traversal attacks where a malicious or misconfigured destination could escape the intended installation directory

## Security Impact
Without this check, a crafted adapter manifest or environment variable could cause the installer to write symlinks to arbitrary filesystem locations. This guard ensures all symlink targets remain within the Claude Code home directory.


rijul-mck (approved):
Reviewed in batch: CI green, mergeable, scanned for owner name, employer and client names, licence-restricted paths, NC licences, credential shapes and self-hosted runner changes.

PR #143, opened 2026-09-06
feat(agents): add prompt injection defense baseline to all core agents

coco-research:
Adds PROMPT-DEFENSE.md and injects it into all 10 core agent system prompts for baseline prompt injection protection.


rijul-mck (approved):

PR #139, opened 2026-09-06
fix(gsd): create missing autonomous workflow and update skill reference

coco-research:
## Summary
- Creates `systems/gsd/workflows/autonomous.md` (233 lines) — the discuss→plan→execute per-phase loop that `gsd-autonomous` SKILL.md referenced but did not exist
- Updates `systems/gsd/skills/gsd-autonomous/SKILL.md` to use repo-relative path `@systems/gsd/workflows/autonomous.md` instead of broken external `@$HOME/.claude/get-shit-done/` reference
- Removes dangling `ui-brand.md` reference

## Testing
- Workflow file validated as syntactically correct markdown with proper frontmatter
- SKILL.md references resolve to existing files within the repository

PR #144, opened 2026-09-06
docs(gsd): document the external GSD dependency and gate both reference forms

coco-research:
This pull request changed direction during this session. An earlier version of it removed 56 references that looked dangling; that was based on a false premise, and this version reverses it. The corrected change documents an external runtime dependency that this bundle has always had, and gates both of its legitimate reference forms so neither can regress silently.

## What was wrong with the previous direction, found and verified in this session

The original commit on this branch rewrote every `@$HOME/.claude/get-shit-done/<rest>` reference in 58 GSD skill files to `@systems/gsd/<rest>`, on the theory that the external path was broken and the repo-relative one would resolve. Measuring the vendored bundle directly disproves that:

```
systems/gsd/ contains only: README.md, agents/, skills/, workflows/
systems/gsd/bin        -> 0 files
systems/gsd/references -> 0 files
systems/gsd/templates  -> 0 files
systems/gsd/workflows  -> 1 file (autonomous.md)
```

`CREDITS.md` describes the vendored scope deliberately and narrowly: "The full GSD orchestration bundle (`systems/gsd/`: 68 skills + 24 agents for phase planning, atomic execution, verification gates)". Skills and agents only. On a machine with GSD installed, `~/.claude/get-shit-done/` exists (`VERSION` 1.34.2) and contains exactly the `bin/`, `contexts/`, `references/`, `templates/`, and `workflows/` directories that this bundle's skill and agent files reference, including every one of the specific files the previous commit's rewrite pointed at instead. The `@$HOME/.claude/get-shit-done/...` references, and the `node "$HOME/.claude/get-shit-done/bin/gsd-tools.cjs" ...` shell invocations elsewhere in these same files, are correct pointers to that separate, intended installation. They are not broken.

Rewriting them to `@systems/gsd/...` is what actually produced the 67-out-of-68 dangling-reference count a prior analysis found: the rewrite pointed at paths that were never vendored. The first version of this branch's fix then made it worse by removing 56 of those references outright, leaving the affected skills with no pointer at all to the workflow they load.

## What this version does instead

- **Restored every one of the 57 affected skill files** to its exact pre-rewrite `@$HOME/.claude/get-shit-done/...` form, verified line-for-line against `62a9c42^` (this branch's own commit that first did the rewrite, at its parent). Restoration was done by reversing the specific token substitution on top of the current merged tree, not by resetting whole files, so every unrelated main-branch edit those files picked up since (for example the frontmatter `description` rewrites from PR #168) is preserved untouched. `gsd-autonomous/SKILL.md` was left alone throughout: its `@systems/gsd/workflows/autonomous.md` reference is genuinely correct, because main commit `efec058` (#139) actually vendored that one workflow file into the repo for real.
- **Re-examined the two previously-flagged "genuine findings" under the corrected understanding.** The one reference that resolved, `workflows/autonomous.md`, is confirmed correct and untouched, for the reason above. The straggler in `gsd-audit-uat/SKILL.md`, a `<process>` sentence with the old `@$HOME/.claude/get-shit-done/workflows/audit-uat.md` form that the original rewrite commit never touched, was added later by PR #168 directly in that form; `~/.claude/get-shit-done/workflows/audit-uat.md` exists, so that reference was correct all along, not a defect. It is left as-is.
- **Documented the dependency in `systems/gsd/README.md`**, in a new section stating plainly what this bundle vendors (skill and agent definitions, plus the one real `workflows/autonomous.md`), what it does not (the upstream framework's workflow, reference, and template documents, and its `gsd-tools.cjs` binary), and that those come from a separate GSD installation expected at `~/.claude/get-shit-done/`, per the upstream project named in `CREDITS.md`. The README's own "Self-contained" claim, which directly contradicted this, was corrected to drop that word.
- **Rewrote `tests/check-gsd-refs.sh`** to check the real invariant rather than one that passed while the defect persisted. It now checks two things: every repo-relative `@systems/gsd/...` reference must resolve to a file this repository actually vendors, and any `$HOME/.claude/get-shit-done/...` reference (which is never itself flagged as dangling) requires `systems/gsd/README.md` to declare that dependency, so an undocumented external path cannot creep back in unnoticed. Wired into `tests/smoke.sh`.

## Verification: every reference classified, not asserted

Every reference under `systems/gsd/` is in exactly one of two legitimate states, or a third, empty category for anything that is neither:

```
State A (repo-relative, resolves inside the repo):                     2
State B (external @$HOME/.claude/get-shit-done, resolves in the
         installed GSD toolkit on this machine):                     156
Neither (a real defect):                                               0
```

State A is `workflows/autonomous.md`'s two citations (its own `execution_context` and `process` lines). State B covers every other citation across `skills/*/SKILL.md` and `agents/*.md`, each checked against the real files present at `~/.claude/get-shit-done/` on a machine with GSD installed, plus the `bin/gsd-tools.cjs` shell invocations, also confirmed present.

## The gate, negative-tested in both halves

Part 1 (repo-relative references must resolve in-repo). Broke `workflows/autonomous.md`'s reference in both lines of `gsd-autonomous/SKILL.md`:
```
  DANGLING (in systems/gsd/skills/gsd-autonomous/SKILL.md): @systems/gsd/workflows/autonomous-BROKEN.md -> systems/gsd/workflows/autonomous-BROKEN.md missing
FAIL: systems/gsd/ references are broken.
      Repo-relative @systems/gsd/... paths must point at a file this
      repo actually vendors. $HOME/.claude/get-shit-done/... paths are
      the legitimate external GSD dependency, but README.md must
      declare it.
```
Reverted (confirmed by an empty `git diff`), then:
```
PASS: systems/gsd/ references are either repo-relative and resolve, or
      external and declared in README.md.
```

Part 2 (README.md must declare the external dependency). Renamed every mention of `get-shit-done` in `systems/gsd/README.md` to something else, leaving the external references in the skill and agent files untouched:
```
  UNDOCUMENTED: systems/gsd/ files reference $HOME/.claude/get-shit-done/
                but systems/gsd/README.md never mentions get-shit-done.
FAIL: systems/gsd/ references are broken.
      ...
```
Reverted (confirmed by `git diff` matching the intended content), then the same `PASS` as above.

## All six required gates pass on this branch, rebased onto fresh `origin/main`

```
python3 scripts/build-index.py        # Done. Skills: 226 · Commands: 44 · Agents: 35
bash tests/check-command-refs.sh      # PASS: all command cross-references resolve.
bash tests/check-evidence-gate.sh     # PASS: /team evidence-gate language present.
bash tests/check-asset-counts.sh      # PASS: all shipped public counts agree with docs/asset-counts.json
bash tests/check-security-surface.sh  # PASS: all security-surface checks passed
bash tests/smoke.sh                   # passed: 26, failed: 0 (includes the rewritten gate)
```

## Files changed

`systems/gsd/README.md` (new dependency section, corrected "self-contained" claim), 57 `systems/gsd/skills/*/SKILL.md` files (restored to their working external-reference form), and `tests/check-gsd-refs.sh` (rewritten to check the real invariant, wired into `tests/smoke.sh`).

### DIFF
--- a/systems/INDEX.md
+++ b/systems/INDEX.md
@@ -10,7 +10,6 @@ Bundles under `systems/` are opt-in via `install.sh --systems <name>`. Directori
 | cognee | `systems/cognee/` | 3 | 0 | 0 | 4 |
 | gsd | `systems/gsd/` | 68 | 24 | 0 | 93 |
 | hyperframes | `systems/hyperframes/` | 20 | 0 | 0 | 928 |
-| learning | `systems/learning/` | 0 | 0 | 0 | 0 |
 | m0 | `systems/m0/` | 4 | 0 | 0 | 9 |
 | superintelligence | `systems/superintelligence/` | 9 | 0 | 0 | 1159 |
 | team | `systems/team/` | 0 | 0 | 0 | 2 |
@@ -20,4 +19,4 @@ Bundles under `systems/` are opt-in via `install.sh --systems <name>`. Directori
 | generic | `adapters/generic/` | 0 | 0 | 0 | 3 |
 | vscode | `adapters/vscode/` | 0 | 0 | 0 | 3 |
 
-> **Advertised as a bundle but installs no artifacts:** `learning`, `team`. These directories hold documentation only, so passing them to `--systems` has no effect.
+> **Advertised as a bundle but installs no artifacts:** `team`. These directories hold documentation only, so passing them to `--systems` has no effect.
--- a/systems/INDEX.md
+++ b/systems/INDEX.md
@@ -10,6 +10,7 @@ Bundles under `systems/` are opt-in via `install.sh --systems <name>`. Directori
 | cognee | `systems/cognee/` | 3 | 0 | 0 | 4 |
 | gsd | `systems/gsd/` | 68 | 24 | 0 | 93 |
 | hyperframes | `systems/hyperframes/` | 20 | 0 | 0 | 928 |
+| learning | `systems/learning/` | 0 | 0 | 0 | 0 |
 | m0 | `systems/m0/` | 4 | 0 | 0 | 9 |
 | superintelligence | `systems/superintelligence/` | 9 | 0 | 0 | 1159 |
 | team | `systems/team/` | 0 | 0 | 0 | 2 |
@@ -19,4 +20,4 @@ Bundles under `systems/` are opt-in via `install.sh --systems <name>`. Directori
 | generic | `adapters/generic/` | 0 | 0 | 0 | 3 |
 | vscode | `adapters/vscode/` | 0 | 0 | 0 | 3 |
 
-> **Advertised as a bundle but installs no artifacts:** `team`. These directories hold documentation only, so passing them to `--systems` has no effect.
+> **Advertised as a bundle but installs no artifacts:** `learning`, `team`. These directories hold documentation only, so passing them to `--systems` has no effect.
--- a/systems/gsd/skills/gsd-add-phase/SKILL.md
+++ b/systems/gsd/skills/gsd-add-phase/SKILL.md
@@ -20,7 +20,7 @@ Routes to the add-phase workflow which handles:
 </objective>
 
 <execution_context>
-@$HOME/.claude/get-shit-done/workflows/add-phase.md
+@systems/gsd/workflows/add-phase.md
 </execution_context>
 
 <context>
@@ -30,7 +30,7 @@ Roadmap and state are resolved in-workflow via `init phase-op` and targeted tool
 </context>
 
 <process>
-**Follow the add-phase workflow** from `@$HOME/.claude/get-shit-done/workflows/add-phase.md`.
+**Follow the add-phase workflow** from `@systems/gsd/workflows/add-phase.md`.
 
 The workflow handles all logic including:
 1. Argument parsing and validation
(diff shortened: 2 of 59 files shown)


rijul-mck (approved):

rijul-mck (approved):
Approving the reversed version. The original premise was wrong and this now does the opposite of what it originally claimed, which is correct.

I verified the inversion independently before this rework was commissioned. systems/gsd/ vendors only a README, an agents directory, a skills directory and one workflow file. The directories the references point into contain zero files, gsd-tools.cjs exists nowhere in the repository, and CREDITS.md scopes the vendoring deliberately to 68 skills and 24 agents. The upstream is installed separately at ~/.claude/get-shit-done/ and does contain all of it. So the external paths were correct pointers to a runtime dependency, and rewriting them to repository-relative form created the 67 dangling references that were then mistaken for the defect being fixed.

The final state is now measured rather than asserted: 2 repository-relative references that resolve in the repository, 156 external references that resolve against the installed dependency, and zero in neither category. The restoration was done by reverse substitution on the merged tree rather than by resetting files, which preserved unrelated edits that had landed on main in the meantime, and it was verified by confirming every pre-rewrite reference line appears verbatim in the restored files.

Two things make this worth landing rather than closing. The bundle README now declares the external dependency explicitly and drops a false self-contained claim, so the next reader does not repeat this diagnosis. And the gate checks the real invariant in two halves, that repository-relative references resolve and that external references require the dependency to be declared, with each half negative-tested separately. The earlier version of that gate validated only one reference form and therefore reported success while 26 binary invocations and 20 document references stayed broken, which is the specific failure mode this replacement avoids.

PR #157, opened 2026-09-06
feat(agent-self-eval): add post-run quality scoring skill and /self-eval command

coco-research:
Adds skills/agent-self-eval/SKILL.md for agent self-evaluation and commands/util/self-eval.md as the /self-eval slash command.


rijul-mck (approved):

PR #164, opened 2026-09-06
feat(goal): add persistent goal skill and /coco:goal command

coco-research:
Adds skills/goal/SKILL.md with continuous-execution loop semantics for goal tracking and commands/coco/goal.md as the dispatcher command for goal management.


rijul-mck (approved):

PR #166, opened 2026-09-12
feat(adapter): add PI-Desktop — one command from clone to in-app

coco-research:
## What

Adds a first-class `pi-desktop` adapter: one command from clone to CoCo running inside PI-Desktop's own UI.

```bash
git clone https://github.com/coco-research/coco.git && cd coco && bash install.sh
```

`install.sh` auto-detects PI-Desktop by checking for its actual roots, `~/.pi/agent` and `~/.agents`, and `npx cocosuperintelligence` forwards to the same path, so the flag is optional. With PI-Desktop already in the adapter list, `--adapter pi-desktop` is explicit when several tools are installed.

## Where it installs

PI-Desktop reads three folders natively. The adapter writes those and nothing else, so CoCo lands in the app's own surfaces, the `/` composer menu, the Skills page, and the Subagents page, with no plugin to install.

| CoCo source | Surface | Destination |
|---|---|---|
| `commands/<ns>/<name>.md` | `/` composer menu | `~/.pi/agent/prompts/<ns>-<name>.md` |
| `commands/<ns>/_index.md` | the bare `/<ns>` | `~/.pi/agent/prompts/<ns>.md` |
| `skills/<name>/` | Skills page | `~/.agents/skills/<name>/` (assets included) |
| `workflows/*.md` | Skills page | `~/.agents/skills/coco-workflow-<name>/` |
| `agents/*.md` | Subagents page | `~/.agents/subagents/coco-<name>.md` |
| `systems/<b>/{skills,agents}` | both pages, with `--systems` | the same two roots |

`--dry-run`, `--uninstall`, `--force` and `--source` are supported.

## Two behaviours worth reviewing

**Front matter is rewritten on the way in.** PI-Desktop caps skill descriptions at 240 characters and several CoCo skills exceed that, so each skill is written with `name` + `description` (+ `domain` where present), clipped on a word boundary. Subagents get `name`, `description`, a tool list (`Read, Glob, Grep, Bash`, plus `Edit` and `Write` for the agents that produce code) and `maxTurns: 80`, the host's own ceiling.

**It will not clobber your own work.** `~/.agents/skills` is shared with hand-written skills, so every file carries a `Coco PI-Desktop adapter` marker and the installer only ever replaces or deletes a file carrying that marker. Anything else is reported and skipped, `--force` overrides, and `--uninstall` removes only its own files.

## Why a script and not a plugin

A PI-Desktop plugin cannot do this job. `manifest.fs.root` accepts only `"workspace"` or `"userSelected"`, so a plugin can reach the workspace or a directory the user picks at runtime, and never `~/.agents` or `~/.pi/agent`. Files are the interface.

## Default scope: an explicit allow-list, and reverse-skill stays opt-in

This adapter changed the framework's default install behaviour: instead of every bundle sitting behind a flag nobody knew to pass, a plain `bash adapters/pi-desktop/install.sh` now installs every bundle in `scripts/installable-bundles.sh`'s list, and `--core-only` is the opt-out. That list used to be computed by scanning `systems/` for any directory that ships a `SKILL.md` or an agent. Once `origin/main`'s `systems/reverse-skill` bundle (a vendored reverse-engineering and penetration-testing methodology pack) merged into this branch, that scan picked it up automatically, which would have put security tooling into every default install with no review at all.

Two things worth stating plainly. First, this was caught before it shipped: on `main` today, `install.sh` sets `SYSTEMS=""` and every adapter's bundle set is empty unless `--systems` is passed explicitly, so nothing installs by default anywhere in the framework right now. This PR's own new default-everything behaviour for PI-Desktop is what would have introduced the exposure, so this is a regression this PR prevented in itself, not a live exposure it fixed. Second, the owner's explicit decision is that `reverse-skill` stays opt-in permanently, independent of how any other bundle's default status is decided.

The fix, in this branch:

- `scripts/installable-bundles.sh` now declares the default set as a fixed, reviewed allow-list, `gsd, brain, cognee, hyperframes, superintelligence, m0`, in one place, instead of deriving it from a directory scan. `reverse-skill` is deliberately absent.
- It stays installable on request through the adapter's own, already-existing `--systems` flag: `bash adapters/pi-desktop/install.sh --systems reverse-skill`. No new flag was introduced; this is the same mechanism every other bundle already used.
- The adapter's own fallback path, which used to re-derive the same directory scan when the shared helper script was missing, now falls back to core-only instead, so the bug cannot resurface there either.
- `install.sh` and `adapters/pi-desktop/install.sh --help` output, and `README.md`, now document `reverse-skill` as opt-in security tooling in full prose, stating plainly that it is reverse-engineering and security-testing material that is not installed unless it is asked for by name.
- A new gate, `tests/check-default-bundle-allowlist.sh`, makes this durable: it requires every directory under `systems/` to have a named, reviewed disposition, so a new bundle can no longer silently join, or silently fail to join, the default set. It is wired into the same verification path as the repository's other `tests/check-*.sh` scripts.

### Proof the gate catches drift

Adding an unreviewed directory under `systems/` fails the gate and names exactly what drifted:

```
=== every systems/* directory has a reviewed disposition ===
  FAIL: systems/drift-test-bundle is not in KNOWN_BUNDLES (tests/check-default-bundle-allowlist.sh) — review it, then decide whether scripts/installable-bundles.sh should add it to the default set

=== Summary ===
  some check(s) failed
```
(exit code 1)

Removing that directory restores a clean pass:

```
=== every systems/* directory has a reviewed disposition ===
  PASS: no undeclared directory under systems/

=== reverse-skill is excluded from the default bundle set ===
  PASS: default bundle set does not include reverse-skill (gsd,brain,cognee,hyperframes,superintelligence,m0)

=== reverse-skill stays reachable through an explicit --systems flag ===
  PASS: adapters/pi-desktop/install.sh --systems reverse-skill selects the bundle

=== Summary ===
  all default-bundle-allowlist checks passed
```
(exit code 0)

### Proof of the installed behaviour

A default install, with no flags, into a throwaway target root installs 178 skills across the six default bundles and no `reverse-skill` content:

```
Installed for PI-Desktop:
  Slash commands : 286   (242 of them Super Intelligence)
  Skills         : 178
  Subagents      : 37
  Bundles        : gsd,brain,cognee,hyperframes,superintelligence,m0
```
No directory under the installed skills path matched `reverse`, `pentest`, `malware`, or `ghidra`.

`--systems reverse-skill`, into a separate throwaway target root, installs it:

```
Installed for PI-Desktop:
  Slash commands : 44
  Skills         : 110
  Subagents      : 10
  Bundles        : reverse-skill
```
14 `reverse-skill`-derived skill directories landed, including `ghidra-reverse`, `pentest-tools`, `malware-analysis`, `apk-reverse`, and `dotnet-reverse`.

## Second defect: the PI-Desktop detection probe never matched

`install.sh`'s auto-detection checked for a directory named `~/.pi-desktop`, but the adapter's own real roots, stated in its own header comment and manifest, are `~/.pi/agent` and `~/.agents`. Nothing creates `~/.pi-desktop`, so that branch could never match on a real PI-Desktop machine and detection silently fell through to the next adapter in the chain. Fixed to check the real roots:

```diff
-  elif [[ -d "$HOME/.pi-desktop" ]]; then
+  elif [[ -d "$HOME/.pi/agent" || -d "$HOME/.agents" ]]; then
     echo "pi-desktop"
```

Verified against a temporary `HOME` in each of three states:

```
=== Test A: neither ~/.pi/agent nor ~/.agents exists ===
Result: generic

=== Test B: ~/.pi/agent exists ===
Result: pi-desktop

=== Test C: ~/.agents exists ===
Result: pi-desktop
```

Neither root present correctly falls through without a false positive; either root present correctly detects PI-Desktop.

## Excluded material

Re-verified after merging `origin/main`, which added `systems/reverse-skill` and `systems/learning` to this branch for the first time: none of the excluded skills (`pwn-chain`, `attack-chain`, `patch-diff-exploit`) or reference files (`frida-bypass-kit.md`, `anti-detection-bypass.md`, `anti-analysis-techniques.md`, `msf-protocol.md`, `network-attack-defense.md`, `web-attack-cheatsheet.md`, `client-side-lab-playbook.md`, `ad-attack-paths.md`) or `pentest-tools/payloads/` are present anywhere in this branch. No absolute local filesystem path was introduced by the merge or by this fix.

## Two more defects, in this PR's own new files

Both of these are in code this PR introduces, not on `main`, so both belong in this change rather than deferred.

**Orphaned dead code in the installer.** `adapters/pi-desktop/install.sh` carried a stray `for`/`return` block inside `front_matter()`, left behind by an incomplete refactor. It sat after that function's own unconditional `return`, so every path through `front_matter` returned before reaching it: unreachable, but syntactically valid Python, which is why it survived undetected. Removed. `bash -n adapters/pi-desktop/install.sh` still passes, and a direct `ast.parse`/`py_compile` of the extracted Python heredoc both still pass. A before/after `diff -r` of a full install into two throwaway target roots (default bundles, and separately `--systems reverse-skill`), one run against the code with the fragment still present and one against the code with it removed, came back byte-for-byte identical in both the file list and the file contents: the deletion changed no behavior.

**Asymmetric smoke coverage.** `tests/smoke.sh`'s adapter dry-run loop covered every adapter except this PR's own `pi-desktop`, while `ci.yml` already covered it: the adapter shipped with smoke coverage in one place and not the other. Added `pi-desktop` to the loop, following the same directory-existence guard already used for `hermes`. `adapters/pi-desktop/install.sh --dry-run` passes cleanly in the loop, and `smoke.sh`'s own pass/fail counter rose from 23/0 to 25/0 (the other new pass is `grok`, added to the loop by a concurrent `main` commit).

## Verification

Rebased onto `origin/main` twice during this work, which in total added the `grok` adapter, widened `tests/check-security-surface.sh` with an unanchored-contains-match-glob check, and corrected site asset counts; both merges were conflict-resolved by hand for prose-only adapter-list lines and by regenerating the generated index files via `python3 scripts/build-index.py`, never by hand-merging them. All green on this branch after the final rebase and every fix above:

```
python3 scripts/build-index.py                  -> indexes current, no drift
bash tests/check-command-refs.sh                 -> PASS: all command cross-references resolve.
bash tests/check-evidence-gate.sh                -> PASS: /team evidence-gate language present.
bash tests/check-asset-counts.sh                 -> PASS: all shipped public counts agree with docs/asset-counts.json
bash tests/check-security-surface.sh             -> PASS: all security-surface checks passed (including the new unanchored-glob check)
bash tests/check-index-tracks-git.sh             -> PASS: generated indexes ignore untracked files.
bash tests/check-default-bundle-allowlist.sh     -> PASS: all default-bundle-allowlist checks passed
bash tests/smoke.sh                              -> passed: 25, failed: 0
```

Both behavioral proofs above were re-run after the final rebase and reproduce exactly: the default install still installs 178 skills across `gsd,brain,cognee,hyperframes,superintelligence,m0` with no `reverse`-matching directory, and `--systems reverse-skill` still installs it (110 skills, `Bundles: reverse-skill`). The excluded-material and absolute-path checks were re-run against the final diff and are still clean.

## Testing

- `ci.yml` now includes `pi-desktop` in the adapter smoke list, and the bundle gate picks the new manifest up automatically (`supports_systems` is exercised for all six bundles).
- A 25-assertion round-trip harness was run against throwaway `HOME`s: install, per-surface content shape, the 240-character budget, subagent name/tool/maxTurns validity, idempotency, the ownership guard, `--force`, `--uninstall`, and that a dry run writes nothing.
- Two defects found and fixed while running it for real, both covered by the harness: `shutil.rmtree` refuses a symlinked skill (the skills directory is shared with linked-in packs, so this would have aborted a run), and the description fallback put a bullet-list fragment in the slash menu for the 21 `team/` commands that carry no frontmatter description.

## Wiring

`install.sh` gains the adapter in its usage block and its detection chain (checked after Claude Code, so no machine that resolves today changes), `ci.yml` gains the smoke entry, `README.md` gains the adapter in the portability blurb, the manual-override list, and a new paragraph documenting `reverse-skill`'s opt-in status, and `systems/INDEX.md` is regenerated with the new row.

## Related

#170 ("install everything by default") carries the same default-scope question from the other direction and is intentionally not touched by this PR; it should be reconciled with the allow-list introduced here once this merges.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,7 @@ jobs:
 
       - name: Adapter smoke test (dry-run)
         run: |
-          for adapter in claude-code cursor vscode codex generic; do
+          for adapter in claude-code cursor vscode pi-desktop codex generic; do
             echo "--- testing $adapter (dry-run) ---"
             bash adapters/$adapter/install.sh --dry-run > /tmp/$adapter.out 2>&1 || {
               echo "FAIL: $adapter"
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ AI chat context is fragile and wipes on `/clear`. CoCo persists phase state, dec
 <td width="33%" valign="top">
 
 ### 3. Vendor-neutral portability
-CoCo compiles its rules, templates, and agent definitions into pure Markdown and YAML frontmatter. Native IDE adapters inject the same library into Claude Code, Cursor, Codex CLI, or any `AGENTS.md` tool — so your workflows follow you even if you switch AI editors.
+CoCo compiles its rules, templates, and agent definitions into pure Markdown and YAML frontmatter. Native IDE adapters inject the same library into Claude Code, Cursor, Codex CLI, PI-Desktop, VS Code, or any `AGENTS.md` tool — so your workflows follow you even if you switch AI editors.
 
 </td>
 </tr>
@@ -566,6 +566,7 @@ bash install.sh --systems superintelligence
 bash install.sh --adapter claude-code
 bash install.sh --adapter cursor
 bash install.sh --adapter codex
+bash install.sh --adapter pi-desktop
 bash install.sh --adapter generic
 ```
 
--- a/adapters/pi-desktop/README.md
+++ b/adapters/pi-desktop/README.md
@@ -1,0 +1,79 @@
+# PI-Desktop adapter
+
+Wires Coco into the three folders PI-Desktop reads natively, so the framework
+shows up in the app's own UI — the `/` composer menu, the Skills page, and the
+Subagents page — with no plugin to install and nothing to drag and drop.
+
+## One command
+
+```bash
+bash install.sh                 # from the repo root; auto-detects PI-Desktop
+bash install.sh --adapter pi-desktop
+```
+
+`npx cocosuperintelligence` works the same way: it forwards to `install.sh`, so
+`npx cocosuperintelligence --adapter pi-desktop` is equivalent.
+
+## Where things land
+
+| Coco source | PI-Desktop surface | Destination |
+|---|---|---|
+| `commands/<ns>/<name>.md` | `/` composer menu | `~/.pi/agent/prompts/<ns>-<name>.md` |
+| `commands/<ns>/_index.md` | the bare `/<ns>` command | `~/.pi/agent/prompts/<ns>.md` |
+| `skills/<name>/` | Skills page | `~/.agents/skills/<name>/` |
+| `systems/<b>/skills/<name>/` | Skills page (with `--systems`) | `~/.agents/skills/<name>/` |
+| `workflows/*.md` | Skills page | `~/.agents/skills/coco-workflow-<name>/` |
+| `agents/*.md` | Subagents page | `~/.agents/subagents/coco-<name>.md` |
+| `systems/<b>/agents/*.md` | Subagents page (with `--systems`) | `~/.agents/subagents/coco-<name>.md` |
+
+Restart PI-Desktop, or open a new session, and the commands and skills appear.
+
+## Options
+
+| Flag | Effect |
+|---|---|
+| `--systems gsd,brain` | Add system bundles. Bundles are opt-in; the core install is skills, commands, agents and workflows. |
+| `--dry-run` | Print the plan and write nothing. |
+| `--uninstall` | Remove every file this adapter wrote, and nothing else. |
+| `--force` | Replace files that are not ours. |
+| `--source <dir>` | Install from a different checkout. |
+
+`PI_AGENT_HOME` and `AGENTS_HOME` override the two target roots, which is what the
+adapter's own tests use.
+
+## Two deliberate behaviours
+
+**It will not clobber your own work.** `~/.agents/skills` is a shared user
+directory — the Skills page puts hand-written skills there too. Every file the
+adapter writes carries a `Coco PI-Desktop adapter` marker, and it only ever
+replaces or deletes a file carrying that marker. Anything else is reported and
+left alone:
+
+```
+Left 1 existing path(s) untouched — not written by this adapter:
+  ~/.agents/skills/my-skill
+Re-run with --force to replace them.
+```
+
+**Descriptions are rewritten.** PI-Desktop caps skill descriptions at 240
+characters, and several Coco skills exceed that. The adapter rewrites each
+front matter block to `name` + `description` (+ `domain` where present), clipping
+the description on a word boundary. Subagents get `name`, `description`, a tool
+list (`Read, Glob, Grep, Bash` for reviewers, plus `Edit` and `Write` for the
+agents that produce code) and `maxTurns: 80`, the host's own ceiling.
+
+## Why this is a script and not a plugin
+
+A PI-Desktop plugin may only reach the workspace or a directory the user picks at
+runtime — `manifest.fs.root` accepts `"workspace"` or `"userSelected"` and nothing
+else — so a plugin cannot write into `~/.agents` or `~/.pi/agent`. Files are the
+interface, and this script writes them.
+
+## Uninstall
+
+```bash
+bash adapters/pi-desktop/install.sh --uninstall
+```
+
+Removes the commands, skills and subagents it created. Skills you wrote yourself
+are left in place.
--- a/adapters/pi-desktop/install.sh
+++ b/adapters/pi-desktop/install.sh
@@ -22,6 +22,7 @@
 #   bash adapters/pi-desktop/install.sh --dry-run              # preview only
 #   bash adapters/pi-desktop/install.sh --uninstall            # remove what it wrote
 #   bash adapters/pi-desktop/install.sh --source /path/to/coco # install a different checkout
+#   bash adapters/pi-desktop/install.sh --force                # replace files this adapter did not write
 #
 # Environment overrides (mainly for tests):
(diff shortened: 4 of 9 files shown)


rijul-mck (approved):

rijul-mck (approved):
Verified independently. The default bundle set is now an explicit allow-list at scripts/installable-bundles.sh line 30, DEFAULT_BUNDLES=(gsd brain cognee hyperframes superintelligence m0), with reverse-skill deliberately absent, and the header comment records why a directory scan was wrong so the decision is not silently undone later.

The gate is scoped to the invariant rather than to one instance, which is what makes it worth having. It checks three separate properties: that every directory under systems/ has a reviewed disposition, so a new bundle cannot join or fail to join the default set unnoticed; that reverse-skill is not in the default set; and that reverse-skill remains reachable through an explicit --systems flag, so the opt-in path is proven rather than assumed.

Both behavioural proofs were reproduced after the rebase: a default install yields no reverse-skill content, and --systems reverse-skill installs it. The pi-desktop probe now tests the real roots at install.sh line 46. I confirmed directly that none of the excluded exploitation material is present in the tree and that the diff contains no absolute user paths.

The dead code removal was verified the right way, by diffing the entire installed tree before and after across 2095 files and finding them byte identical, so the deletion is proven behaviour preserving rather than argued to be. Smoke coverage for the new adapter rose from 23 to 25 checks.

One cosmetic item is deliberately not blocking this: a few newly authored comment lines use an em dash, which the house style disallows. That will be cleaned up in a single sweep rather than costing this change its own serial integration run.

PR #167, opened 2026-09-12
fix(index): generate from git-tracked files, not the working tree

coco-research:
## The bug

Every generated index was a function of the **working tree**, not of the commit. `build-index.py` enumerated with `Path.glob` / `iterdir` / `rglob`, so anything sitting on disk got indexed as if it shipped.

That is the root cause of the incident that cost twenty PRs. A stale, untracked `systems/<name>/` directory produced a phantom bundle row in `systems/INDEX.md`, and every branch carrying that regenerated index then failed the *Verify INDEX is up to date* gate with drift its author had not caused and could not explain. The same mechanism let an untracked `skills/<name>/` reach `skills/INDEX.md`.

## The fix

Discovery filters through `git ls-files`. Four call sites changed: the four `SKILL_SOURCES` globs, `commands/*/*.md`, both agent globs, and the `systems/` + `adapters/` container walk (including its file count, so an untracked directory cannot pad a bundle's totals either).

- **On a clean checkout the output is byte-identical** to the previous generator — verified by hashing all ten generated files before and after.
- **Outside a git checkout** (a tarball or npm install) it falls back to the filesystem walk and prints a warning on stderr, because there is nothing else to read.

## The test

`tests/check-index-tracks-git.sh` plants an untracked `systems/` and `skills/` probe, regenerates, and fails if either appears. Against the **previous** script it fails on all three assertions; against this one it passes. It runs in the standalone-suite step, and cleans up via a trap so that step's dirty-tree check still holds.

Proven both directions before committing:

```
# previous script
FAIL: an untracked systems/ directory changed systems/INDEX.md
FAIL: an untracked skills/ directory changed skills/INDEX.md
FAIL: the probe was indexed

# this one
PASS: generated indexes ignore untracked files.
```

## Why it matters beyond the incident

The index is the repo's published inventory of its own skills and agents. A generator that indexes whatever is on disk will eventually assert that a bundle exists when it does not — and the CI gate that compares generated output cannot tell the difference between "the index is stale" and "the tree is dirty". This removes that ambiguity: the index is now a pure function of the commit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -152,12 +152,15 @@ jobs:
 
       - name: Run the standalone test suites
         run: |
-          # These three suites pass locally but were invoked nowhere in CI, so nothing stopped
+          # These suites pass locally but were invoked nowhere in CI, so nothing stopped
           # them rotting. m0-smoke.sh and run_fixtures.sh had no CI coverage at all.
           bash tests/smoke.sh
           bash skills/arch-index/scripts/run_fixtures.sh
           bash systems/m0/tests/m0-smoke.sh
-          # tests/smoke.sh writes into the tree; fail loudly rather than leaving it dirty.
+          # The generated indexes must not depend on untracked files: a stale
+          # systems/<name>/ directory used to produce a phantom bundle row and show
+          # up as unexplained INDEX drift on twenty unrelated branches.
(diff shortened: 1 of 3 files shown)


rijul-mck (approved):

PR #168, opened 2026-09-12
docs(descriptions): say when to use each file, not just what it is

coco-research:
## The defect

A `description` is the only signal a loader has for whether to open a file. 91 of these 140 never named a situation that should trigger one: they described their subject and stopped. The skill then never fires when it is actually needed — the audit's most expensive finding, and the one no CI gate can see.

```
before  "Expert in shipping production-grade AI-powered features — LLM integration
         patterns, RAG architecture, prompt engineering that scales ..."   (269 chars)
after   "Use when building AI features into a product: LLM integration, RAG pipelines,
         guardrails, streaming, AI UX, prompt engineering, or AI cost control."
```

## What changed

Each description now **leads with the trigger clause**, names the user's own phrasings, artifacts or file types, and stays inside the **240-character cap** PI-Desktop applies to its catalog. Detail that crowded the description belongs in the body, which is read once the skill loads.

Measured over the same 140 files with the scanner that found the defects:

| | no trigger | over 240 | no description |
|---|---:|---:|---:|
| before | 91 | 14 | 41 |
| after | **0** | **0** | **0** |

- **23 files had no front matter at all** (all under `commands/`), which is why 23 rows of the slash menu rendered blank. They now carry a description.
- One file, `rules/cursor-mdc/agents-md-creation.mdc`, had unquoted globs that broke `yaml.safe_load`; it is valid now.

## Body fixes: the 37 entries the audit marked broken

Scope was held to the issue recorded for each file, and nothing else:

- `verification-agent`, deleted from the repo, is no longer invoked; the verdict template it referenced is inlined.
- `brain-wiki`'s phantom `brain_cli.py wiki --list-merges` / `--approve-merge` subcommands are replaced with the direct `knowledge.db` access that actually exists.
- `media-memory` no longer claims Gemini Embedding — its own body says it uses local all-MiniLM-L6-v2 vectors — and its non-standard `trigger:` key is folded into the description.
- `commands/design/mermaid.md` drops a `~/Downloads/...` dependency for the package version its own templates already pinned, and its unsubstituted `{{ARGUMENTS}}` becomes `$ARGUMENTS`.
- `gsd-review-backlog` no longer instructs the reader to rename a directory the preceding step had just created; it moves the contents instead, and the delete step now requires a second confirmation before `rm -rf`.
- Stale `(FIX Cx/Mx)` review-finding citations are removed from user-facing text.

## Verification

- Frontmatter lint: **3/3**. Full `ci.yml` mirror: **11/11 steps**, including the *Verify INDEX is up to date* gate.
- Every description is single-line, double-quoted and YAML-valid.
- A key-preservation check across all 140 files confirms **no file lost a front matter key** and no new YAML error was introduced (one pre-existing error was fixed).
- `skills/INDEX.md` and `docs/by-domain/*.md` regenerated, so the published catalog matches.

## Note for review

`commands/**` is not yet covered by the frontmatter lint — it validates `skills/*/SKILL.md`, `systems/**/SKILL.md` and `agents/*.md` (widened in #165). The 23 new front matter blocks are verified by hand here; extending the lint to commands is a sensible follow-up.


rijul-mck (approved):

PR #169, opened 2026-09-12
feat(index): record what each adapter actually delivers, and rebuild the stale arch index

coco-research:
## Why this exists

The repo could describe its assets but not its install surface, and those differ by a factor of eight.

`commands/` contains **38** files. The README advertises **280 commands**. The other 242 are generated at install time by `systems/superintelligence/ai/scripts/build_commands.py` (225 per-team) and `systems/superintelligence/scripts/build_meta_commands.py` (17 cross-team). An adapter that does not invoke those generators delivers 38 commands and says nothing about it, which is exactly how a PI-Desktop install ended up missing 87% of the command surface, and why the whole problem was invisible to the audits, to CI, and to anyone reading the tree.

## What it adds

`scripts/build-delivery-index.py` answers "what does an install deliver" the only way that cannot drift: **it runs the installers.** Each adapter is installed into a throwaway `HOME` at its widest advertised configuration, and the result is counted. Counts follow symlinks and deduplicate by resolved path, because most adapters link into the checkout rather than copying; and they are scoped to the adapter's own target roots, so a skill that ships its own `commands/` directory (visual-explainer has seven) is not mistaken for delivered commands.

The first run, which is the diagnosis:

| Adapter | Commands | of which generated | Skills | Agents | AGENTS.md |
|---|---:|---:|---:|---:|:--:|
| `claude-code` | **280** | 242 | 171 | 34 | — |
| `vscode` | **280** | 242 | 171 | 34 | — |
| `cursor` | **38** | 0 | 75 | 0 | — |
| `codex` | 0 | 0 | 0 | 0 | yes |
| `generic` | 0 | 0 | 0 | 0 | yes |

Cursor is **242 commands short**, and it is short because its manifest declares `supports_systems: []`, a deliberate claim that the existing bundle gate enforces, not an oversight. Cursor cannot install a bundle even when asked. Codex and generic are AGENTS.md producers by design and are reported as not comparable rather than as failures.

The generated file also states the measurement's own conditions: the VS Code profile is created inside the throwaway HOME because its core commands need one, and skills stop at 171 rather than the published 185 because five of those skills live under `adapters/cursor/skills` and are Cursor-only.

## Gating

`adapters/INDEX.md` is committed and a new CI step regenerates it and fails on any diff, so an adapter that stops delivering a surface becomes a reviewable diff instead of a user noticing months later that most of the framework never arrived. The step takes about nine seconds.

## Also: the architecture index was rebuilt

`.arch/` had drifted twice over. Its two artifacts disagreed with **each other**: `index.json` claimed `e5c673f3` (generated 2026-07-31) while `.arch/pinned-commit` held `3dc71a3`, and neither matched HEAD. The deterministic scan put the drift at `FULL_REBUILD`: **533 files and 65,521 lines** since the pin.

Rebuilt by crawl (`scripts/repo_tree.py`, depth 4, 706 files, not truncated), per `skills/arch-index/SKILL.md`. All five components survived reconciliation, so no identifier changed. Two purposes were refreshed where they had gone stale:

- **install-adapters** now names its five targets (Claude Code, Cursor, VS Code/Copilot CLI, Codex, generic AGENTS.md) rather than three.
- **command-surface** now records the fact that matters most about it: that 242 of its commands are generated at install time rather than committed, so the installed surface is roughly eight times the size of the directory.

The validator exits **0 with 12/12 checks and 10/10 paths verified**, so the pin advanced per step 5 of the skill. `.arch/ARCH-EVIDENCE.md` captures the gate; `.arch/INDEX.md` points at `adapters/INDEX.md` for the delivery question the structural index deliberately does not answer.

## Verification (original)

- Full `ci.yml` mirror: **12/12 steps pass**, including the new delivery-index gate, the INDEX freshness gate, and the bundle install gate.
- `validate_index.py`: exit 0, 12/12 checks, 10/10 paths verified, 0 violations.
- The delivery index was checked against a hand count: `claude-code` reports 280 commands, which is the README's figure, and its 242 generated commands match the generators' own output exactly.

## Correction: the environment pop list and the advertised totals were themselves stale snapshots

An independent review of this branch found that both of the defenses this pull request added were themselves hardcoded snapshots of a smaller tree, and both had already gone stale by the time of the review. This section documents what was wrong, what changed, and the evidence for the fix. The claim above that "each adapter is installed into a throwaway `HOME`" and writes nowhere else did not hold for most adapters, for the reason explained below; it holds now.

**The environment variable pop list.** At the time this pull request was opened, `scripts/build-delivery-index.py` stripped exactly nine variables before invoking an adapter's installer: `CLAUDE_HOME`, `CURSOR_HOME`, `COPILOT_HOME`, `PI_AGENT_HOME`, `AGENTS_HOME`, `XDG_CONFIG_HOME`, `XDG_DATA_HOME`, `APPDATA`, and `LOCALAPPDATA`. That list was correct when the repository had five adapters. Once `origin/main` was merged into this branch, the repository has fifteen adapters, and eleven of them read their own environment variable to redirect where they write, following the convention `${SOME_VAR:-$HOME/...}`: `AMAZON_Q_HOME`, `CLAUDE_HOME`, `CLINE_HOME`, `CONTINUE_HOME`, `COPILOT_HOME`, `CURSOR_HOME`, `GROK_HOME`, `HERMES_PROFILES`, `ROO_CODE_HOME`, `WINDSURF_HOME`, and `ZED_HOME`. Only three of those eleven were in the original list. Counting the two platform fallback variables the VS Code adapter also reads, `APPDATA` and `XDG_CONFIG_HOME`, thirteen real override variables existed in total, and the original nine-entry list covered five of them while leaving eight unstripped. Four of the original nine entries, `PI_AGENT_HOME`, `AGENTS_HOME`, `XDG_DATA_HOME`, and `LOCALAPPDATA`, are read by no adapter anywhere in the tree and were dead weight from the start. In practice, this meant that running the delivery index build on a machine whose ambient shell happened to export, for example, `ROO_CODE_HOME` or `HERMES_PROFILES`, such as a developer's own profile or a CI runner's configuration, would not have been confined to the throwaway `HOME` this script constructs. It would have written real files into whatever real path that variable pointed at.

**The advertised totals.** The `ADVERTISED` dictionary was the same kind of snapshot: `{'commands': 280, 'skills': 185}`, both copied from the README at the time. `docs/asset-counts.json`, the generated source of truth that `scripts/build-index.py` produces from a full walk of the tree and that `tests/check-asset-counts.sh` already gates against the README, now reports 222 skills, a customer-facing total of 286 commands, and a shipped baseline of 44 commands, all higher than the numbers this pull request had hardcoded.

**The fix.** Both were changed from a hand-kept list to a derivation. `env_override_vars()` now scans every file matching `adapters/*/install.sh`, finds every occurrence of the pattern `${SOME_VAR:-$HOME/...}` (the braced `${HOME}` form is also recognized), and returns the variable names it finds. A newly added adapter that follows the repository's own convention has its override variable captured automatically, the moment its `install.sh` lands, without anyone editing this script. `advertised_totals()` now reads `commands.customer_facing`, `commands.shipped`, and `skills.total` directly out of `docs/asset-counts.json` instead of holding a copy of them. The three prose sentences inside the generated `adapters/INDEX.md` that used to quote the old 280, 185, and 38 figures verbatim now interpolate the same live values, so the generated file cannot contradict itself between its table and its footnotes the way a second hardcoded copy would eventually have caused it to.

**Isolation proof.** `ROO_CODE_HOME`, one of the eight variables the original list left unstripped, was set to a real directory outside any throwaway `HOME`, and `install('roo-code', ...)` was called through the fixed script with that variable present in the ambient environment. The target directory remained completely empty afterward: `find "$MARKER" -mindepth 1` produced no output and exited 0. As a control, the same test was run with `CLAUDE_HOME`, one of the three variables the original list already stripped, pointed at a second real directory; that directory also stayed empty, showing the fix preserves the behavior that was already correct for those three. To confirm `ROO_CODE_HOME` is a genuine write vector rather than a vacuous test, `adapters/roo-code/install.sh` was then run directly, bypassing the fixed script's stripping entirely, with `ROO_CODE_HOME` pointed at a third real directory and the real `HOME` left untouched. That run wrote real symlinks into it, including the line `Linked: .../marker-roo-leak/agents/ui-ux-designer.md -> .../agents/ui-ux-designer.md`, confirming the variable really is the kind of escape route the pop list exists to close.

**Drift proof.** A new adapter directory, `adapters/faketool/`, was created containing only an `install.sh` whose only relevant line was `TARGET_HOME="${FAKETOOL_HOME:-$HOME/.faketool}"`. Before that directory existed, `env_override_vars()` returned thirteen variables. Immediately after, without any change to `scripts/build-delivery-index.py`, it returned fourteen, and the new entry was exactly `FAKETOOL_HOME`. The directory was then deleted, and the function returned to the same thirteen variables it started with. The same style of proof was run for the advertised totals: a new skill directory, `skills/zzz-drift-test-skill/`, was staged with `git add` (the generator only counts git-tracked files, by design, per `tests/check-index-tracks-git.sh`), and regenerating `docs/asset-counts.json` moved its reported skill total from 222 to 223; `advertised_totals()` picked up 223 immediately, with no code change. The skill directory was then unstaged, removed, and the index regenerated again, returning the total to 222.

All six required gates pass on this branch after the fix and after merging `origin/main`: `python3 scripts/build-index.py`, `tests/check-command-refs.sh`, `tests/check-evidence-gate.sh`, `tests/check-asset-counts.sh`, `tests/check-security-surface.sh`, and `tests/smoke.sh` (24 passed, 0 failed). The one merge conflict, in `.github/workflows/ci.yml`, was between this branch's new "Verify the delivery index matches what the adapters install" step and `origin/main`'s new "Verify shipped asset counts match generated truth" step; both are independent gates and both are now kept, one after the other.

## Second correction: main moved again, and the architecture index conflicted with itself

`origin/main` advanced past the merge above almost immediately, landing a Grok department-wiring PR, a `pi-desktop` adapter, GSD dependency documentation, and an adapter-table update, sixteen adapters in total. Merging that fresh tip produced one genuine new conflict, in `.arch/ARCH-EVIDENCE.md`, `.arch/INDEX.md`, `.arch/index.json`, and `.arch/pinned-commit`, plus a clean, non-conflicting merge everywhere else, including `.github/workflows/ci.yml` (the adapter dry-run loop there is still the hardcoded, pre-#180 list on `main` as of this push; no second conflict has occurred there yet).

**What `.arch/index.json` actually is.** It is not a script's deterministic output. `skills/arch-index/SKILL.md` step 2 is "synthesise, with no tool use": a model reads a tree crawl and writes the component map by judgment, then a validator (`validate_index.py`) checks the result mechanically. `.arch/INDEX.md` and `.arch/ARCH-EVIDENCE.md` are, by contrast, pure renderings of that JSON plus the validator's and drift scanner's own output, and `.arch/pinned-commit` is `git rev-parse HEAD` written only after the validator exits zero. The conflict was between two independent syntheses: this branch's own rebuild and a separate rebuild that had landed on `main`, both pinned to the same upstream commit, `992abf3489adf5bb229b17a4a3036cd1690795f5`, but written with different prose and, on `main`'s side, a `pinned-commit` file missing its trailing newline. Neither side's JSON was hand-merged. Instead, `python3 skills/arch-index/scripts/arch_drift.py` was run against the fully merged tree from that shared pin: `gate=FULL_REBUILD`, `filesChanged=552`, `linesChanged=46670`, and, critically, all five existing components (`skill-library`, `command-surface`, `system-bundles`, `agent-roster`, `install-adapters`) came back `KEEP` with every one of their claimed directories still present. The scale-up was entirely growth inside already-claimed directories, ten new adapters under `adapters/`, seven new Super Intelligence departments under `systems/superintelligence/`, so no component boundary changed and no sixth component was warranted; the new `mcps/` reference-documentation directory and the rest of the forty unclaimed files stay unclaimed under the runtime-only and anti-pattern rules in `references/component-rules.md`, the same rule that already excluded `scripts/`, `docs/`, `tests/`, and `.github/`. A fresh `index.json` was synthesized against the merged tree, reusing the accurate parts of both prior versions and refreshing the parts that had gone stale, `validate_index.py` returned exit 0 with 12/12 checks and 10/10 paths verified, and `pinned-commit` was advanced to `6a22fe8a72bcb848263385f80a6bd22089fe9173`, `main`'s tip at merge time, matching how the prior pin was itself set to `main`'s tip rather than to any branch's own merge commit.

**Re-derived totals, and a live example of the drift this fix is for.** After this second merge, `docs/asset-counts.json` reports 226 skills and a customer-facing total of 386 commands, both higher again than the 222 and 286 measured after the first merge; `advertised_totals()` picked up both with no code change. More directly relevant: the new `pi-desktop` adapter reads both `PI_AGENT_HOME` and `AGENTS_HOME`, two of the four variables that were dead weight in this pull request's original nine-entry pop list. `env_override_vars()` now returns fifteen variables, including both, again with no code change, exactly the property the derivation exists to provide. All seven required gates were re-run after this merge, including the newly-added `tests/check-default-bundle-allowlist.sh`, and all pass: `python3 scripts/build-index.py` (226 skills, 44 shipped commands, 35 agents), `tests/check-command-refs.sh`, `tests/check-evidence-gate.sh`, `tests/check-asset-counts.sh`, `tests/check-default-bundle-allowlist.sh`, `tests/check-security-surface.sh`, and `tests/smoke.sh` (26 passed, 0 failed).

Both proofs from the first correction were re-run against the sixteen-adapter tree rather than trusted from before. The isolation proof was repeated with `PI_AGENT_HOME` set to a real directory outside any throwaway `HOME`, calling `install('pi-desktop', ...)` through the fixed script; that directory stayed empty, and a `CLAUDE_HOME` control also stayed empty. Run directly, bypassing the fix, with `HOME` pointed at a scratch directory and `PI_AGENT_HOME` pointed at a second real directory outside it, `adapters/pi-desktop/install.sh` wrote all 386 of its command files into that second directory, confirming the vector is real for this adapter too. The drift proof was repeated with a second throwaway adapter directory: `env_override_vars()` went from fifteen variables to sixteen the moment the scaffold's `install.sh` existed, with zero code change, and back to fifteen the moment it was removed.


rijul-mck (approved):

rijul-mck (approved):
Verified. The evidence behind this one is the strongest in the batch, because the vector was demonstrated rather than argued.

The environment pop-list was a snapshot taken when the repository had five adapters. Measured against the current tree, 8 of 11 target-root override variables went unstripped, and 4 of the 9 entries it did carry are read by no adapter at all. The counterfactual settles that this mattered: with the stripping bypassed, an override pointed outside the sandbox caused real files to be written to that location, so a delivery-index run genuinely could have written into a live configuration root.

The fix derives the list by scanning each adapter's own installer for the override convention rather than extending the list by hand, and the drift proof shows a newly added adapter's variable picked up with no code change. The advertised totals are likewise read straight from the generated source of truth and now report the current 226 and 386 with no stale figure hardcoded anywhere. That is the correct shape: the enumeration is no longer a second copy of a fact the tree already knows.

The pull request's original environment-isolation claim was false for those 8 adapters, and the body now says so plainly instead of repeating it. Correcting an overclaim rather than quietly shipping it is the right call.

The architecture index conflict was handled carefully. Rather than hand-merging two independent syntheses, the drift tool was run against the merged tree from the shared pin, all five components verdicted unchanged with every claimed directory still present, and a fresh index synthesised and validated at 12 of 12 checks before advancing the pin.

One operational note, recorded because it is worth remembering rather than because it affects this diff: while re-running the isolation proof, the test itself wrote 386 files into a real tool directory because HOME was not overridden alongside the adapter's own variable. That was disclosed rather than hidden, the remaining proofs were re-run correctly, and the incident is an instance of exactly the class of leak this change prevents. The affected directory has been verified as entirely created by that run and is being restored.

PR #126, opened 2026-08-30
feat(adapters): add Grok Build / Grok CLI adapter

coco-research:
## Summary

- Adds a first-class `grok` adapter that wires Coco into `~/.grok/{skills,commands,agents,rules,hooks}`, matching the Claude Code / Cursor / VS Code adapters.
- Commands flatten to `<ns>:<name>.md` because Grok discovers user commands as flat `*.md` files and does not recurse. Rules are symlinked from Cursor `.mdc` sources as `*.md`.
- `--systems` is supported, including Super Intelligence (`si-<team>` skills + 242 generated `/SI-*` commands). When `coco-platform-mcp` is present it is registered in `~/.grok/config.toml` so native Grok config wins over Cursor's relative `./backend` MCP entry; m0 hooks are written when `coco-m0-hook` is present.
- Root installer auto-detects `~/.grok` after VS Code. Smoke suite and CI dry-run loop cover the new adapter.

## Test Plan

- [x] `bash adapters/grok/install.sh --dry-run` (and `--systems gsd`)
- [x] `bash install.sh --list` shows `grok`
- [x] `bash tests/smoke.sh` adapter dry-run loop includes grok and passes
- [x] `bash -n adapters/grok/install.sh` and `bash -n install.sh`
- [ ] CI adapter dry-run job on this PR
- [ ] Fresh Grok session sees skills/commands from `~/.grok/` and `coco-platform` MCP from `config.toml`

PR #170, opened 2026-09-12
fix(install): every bundle by default except security tooling, which stays opt-in

coco-research:
## The corrected intent

This pull request originally claimed "everything by default," including the reverse-engineering and security-testing bundle. That claim is now false: `main` has since gated the default install to an explicit, reviewed allow-list in `scripts/installable-bundles.sh`, deliberately excluding `systems/reverse-skill`, and `tests/check-default-bundle-allowlist.sh` enforces that exclusion. "Everything by default" was the wrong premise from the start, because it would have shipped penetration-testing and reverse-engineering tooling to every user who ran a plain install with no flags. The corrected intent is that every bundle in the reviewed allow-list installs by default, and the security-testing bundle stays opt-in via `--systems reverse-skill`.

This pull request has been amended to build on that decision rather than around it. The default bundle set still comes from `main`'s allow-list; nothing here reintroduces a scan of `systems/`, which is exactly the defect the allow-list was written to prevent (a scan would silently add `reverse-skill` to every default install the moment the directory exists).

## What this pull request actually changes, once reconciled

Its real, surviving contribution is wiring the adapters to the allow-list, not widening the allow-list. Before this change, a plain `bash install.sh` on every adapter except `pi-desktop` installed the core set only, because bundles required an explicit `--systems <list>` that nobody knew to pass. That is still true on `main` today for `claude-code`, `cursor`, `vscode`, `codex`, and `generic`. This pull request:

- Adds `--core-only` and default-bundle resolution (via `scripts/installable-bundles.sh`) to the top-level `install.sh` and to the `claude-code`, `cursor`, `vscode`, `codex`, and `generic` adapters, matching the pattern `main`'s `pi-desktop` adapter already established.
- Extracts `scripts/generate-si-commands.sh` as the one shared place that runs the Super Intelligence command generators, so `claude-code` and `cursor` no longer carry their own duplicate inline implementation of that step.
- Brings the Cursor adapter to parity: it now installs bundle skills and generates the SI family, and its manifest declares the six bundles it genuinely supports.
- Fixes 59 stale persona file paths in the AI team registry (`superintelligenceTeam/...` pointing at a directory that no longer exists).

Superseded by `main` and discarded from this pull request: the copy of `scripts/build-delivery-index.py` this branch carried is gone, replaced wholesale by `main`'s version from #169. That version derives the environment-override list and the advertised totals from the tree instead of hardcoding them; the old hardcoded override list had left 8 of 11 variables unstripped, so a delivery-index run could write into a live configuration root. One further correction was made on top of `main`'s version: the generated `adapters/INDEX.md` carried a note claiming "Cursor declares `supports_systems: []`," which this pull request's own Cursor-parity work makes false, so that stale bullet is removed.

## Corrected figures

Every number this pull request stated was checked against `docs/asset-counts.json`, the generated source of truth, and corrected where stale (mostly because four more departments became built teams since this branch was opened):

| Claim | Was | Now |
|---|---:|---:|
| Skills | 185 (70 core + 115 bundle) | 226 (74 core + 152 bundle) |
| Slash commands | 280 (38 core + 242 generated) | 386 (44 core + 342 generated) |
| Specialized agents | 37 (10 core + 27 bundle) | 35 (11 core + 24 bundle) |
| Expert personas | 389 | 495 |
| SI command generators | 225 per-team + 17 cross-team (9 team registries) | 325 per-team + 17 cross-team (13 team registries) |

## Behavioral proof

Both runs below used a scratch `HOME` and every adapter-specific override variable set to a scratch path (`CLAUDE_HOME`, `CURSOR_HOME`, `COPILOT_HOME`, `PI_AGENT_HOME`, `AGENTS_HOME`, `HERMES_PROFILES`, `XDG_CONFIG_HOME`, `XDG_DATA_HOME`, `APPDATA`, `LOCALAPPDATA`), so nothing could fall back to a real home directory.

**Default install, no flags:**

```
Installed for Claude Code:
  Slash commands : 386   (44 core + 342 Super Intelligence)
  Skills         : 175
  Subagents      : 35
  Bundles        : gsd,brain,cognee,hyperframes,superintelligence,m0
```

The bundles installed are exactly the allow-list. A search of the installed tree for `reverse-skill`, `pentest`, `malware`, `attack-chain`, and `pwn-chain` by name, and of every symlink's resolved target for `reverse-skill` or `pentest-tools`, returned zero matches.

**Explicit opt-in, `--systems reverse-skill`:**

```
Installed for Claude Code:
  Slash commands : 44   (core commands only)
  Skills         : 107
  Subagents      : 11
  Bundles        : reverse-skill
```

The security bundle installed this time, including `reverse-engineering`, `radare2`, `thick-client`, and `threat-hunting`, confirming the flag actually selects it. In both runs, nothing was written outside the scratch `HOME`: the working directory the installer ran from stayed empty, and the real `~/.claude/skills` directory received no new files.

## Verification

Merged fresh `origin/main` (`#126`, `#166`, `#138`, `#144`, `#120`, `#171`, `#169`, `#177`, `#178`, `#179`, `#180`, `#181` all included). All generated files (`skills/INDEX.md`, `commands/INDEX.md`, `agents/INDEX.md`, `systems/INDEX.md`, `docs/by-domain/*.md`, `docs/asset-counts.json`, `adapters/INDEX.md`) were regenerated rather than hand-merged. All eight required gates pass: `build-index.py`, `check-command-refs.sh`, `check-evidence-gate.sh`, `check-asset-counts.sh`, `check-security-surface.sh`, `check-default-bundle-allowlist.sh`, `check-adapter-drift.sh`, and `smoke.sh`.


rijul-mck (approved):

rijul-mck (approved):
Approving the amended version. As originally written this could not merge, because its premise contradicted a decision that has since landed.

The title now reads that every bundle installs by default except the security tooling, which is the accurate description. I verified the substance rather than the wording. scripts/installable-bundles.sh still declares the explicit allow-list, DEFAULT_BUNDLES=(gsd brain cognee hyperframes superintelligence m0), with no directory scan of systems/ reintroduced. That matters because the scan is precisely what would have added the vendored reverse engineering bundle to every default install without review. reverse-skill remains absent from the default set, and the three remaining mentions in that file are the comments explaining why. Running tests/check-default-bundle-allowlist.sh against this branch passes all its checks, including the one proving the bundle is still reachable through an explicit flag.

The delivery index script is one hunk away from main's version, and that hunk is a correction rather than a regression. It removes a note asserting that Cursor declares no system support and can therefore never reach the Super Intelligence family. I checked the manifest on both sides: Cursor already declares all six bundles on main, so that note was describing a state the repository had already left, and deleting it removes a false claim. The rest of main's rewritten script, where the environment override list and the advertised totals derive from the tree, is preserved intact.

The stale figures are gone. Nothing in the diff reintroduces the superseded command, skill or agent totals, which matters because several gates now enforce published claims against the generated source of truth, including prose claims in the README.

The adapter entry points are consistent with the amended intent: each defaults to the reviewed allow-list, with an override flag to narrow the selection and a core-only flag to install no bundles at all. None of the excluded exploitation material returned, and the diff contains no absolute user paths.

PR #172, opened 2026-09-14
fix(si): keyword Stage A fallback + COCO_SI_REPO on generate

coco-research:
## SUCCEEDS-WHEN (Kavya eng + Mira product)
1. `meta_select.py` exits 0 with `method: keyword|embed` when LM Studio `:1234` is down/up.
   On keyword: **supported degraded production path** — `delegate: false`, stderr WARN, `degraded: true`.
   Do not reuse embed FLOOR. Do not claim keyword ≡ embed. Fail closed if keyword also cannot run.
2. `adapters/claude-code/install.sh` sets `COCO_SI_REPO` when generating SI commands (Decide B eng; not a branding change).

## Product call (Mira, written)
Keyword fallback = supported degraded production path — **not** dogfood-only unsupported.
User-facing honesty: say degraded / keyword fallback; never silent full SI.

## Evidence
Ash dogfood: `:1234` ConnectionRefused → exit 0, `method:keyword`, `delegate:false`, stderr WARN.

## Out of scope
#122 / #124 untouched. No site.


rijul-mck (approved):

PR #125, opened 2026-08-30
[site] hide Platform and PDF from public catalog

coco-research:
## Summary
- Remove CoCo Platform and CoCo PDF from the public catalog (products overlay + homepage tiles). Platform is not public; PDF is unpromoted.
- Update count copy so the page does not still say fifteen/fourteen after those two are gone (thirteen things, twelve more; framework tab 3; desktop-apps tab 3).
- No flagship CTA change, no install-command change, no README/logo/skill edits.

## Test plan
- [ ] Overlay has no CoCo Platform or CoCo PDF
- [ ] Homepage tiles have no `#product-platform` or `#product-pdf`
- [ ] Remaining products still listed
- [ ] Count copy matches 13 public items
- Pike reviews. Mira signs off copy.

PR #173, opened 2026-09-14
[site] flagship CTA: bash install.sh --systems superintelligence

coco-research:
## Summary
- Flagship install CTA on `/coco/` is now `bash install.sh --systems superintelligence` (Annu: #124 is on main).
- Minimal surrounding install prose updated so the page does not still describe the old bootstrap one-liner.
- Counts and version untouched (185/280, v1.2.0). No homepage catalog changes. No README.

## Test plan
- [ ] Install step one shows the exact Superintelligence install command
- [ ] Step three still `/SI-Decide`
- [ ] Specs table still 185/280 and 1.2.0
- Mira signs copy. Pike reviews. Rijul merges.


coco-research (commented):
Site-lane review (Pike). I cannot approve or merge as coco-research (same actor as the author).

CTA change: LGTM in intent. Step one → `bash install.sh --systems superintelligence` matches Annu/#124-on-main and the earlier sales ask. Counts/version not touched. #124 and #125 are on main.

BLOCKED on scope/regression in the same diff:
- Removes `prefers-reduced-motion` handling for scroll + overlay (and the reduced-motion reveal/overlay/nav exceptions beyond a stub).
- Removes `:focus-visible` ring.
- Shrinks nav / nav-trigger / overlay-close hit targets (drops min-height 44px / padding).
- Removes overlay focus-on-open, focus-restore-on-close, and Tab trap — the S-M3 class of work from the a11y lane.
- Videos switched to `loop autoplay` and drop `controls` + the short fine captions.

Those are not required for the install CTA. Ship the CTA + install prose alone, or restore the a11y/CSS/JS behavior in a follow-up before merge.

Mira still signs copy. I will not merge.

coco-research:
Fixed. First push was against a stale `coco/index.html` and looked like an a11y regression.

Branch reset to current main. Diff is now **install CTA only** (+5/−6 in the install section). a11y from main kept (reduced-motion, focus-visible, 44px targets, overlay focus/trap, video controls). Please re-review.

PR #175, opened 2026-09-14
fix(hermes): create the target profile instead of failing the install

rijul-mck:
## Summary

`adapters/hermes/install.sh` defaults to the `dev` profile and treated a missing profile directory as fatal, so a **first install failed on any machine where no Hermes profile had been created by hand**.

The delivery gate from the index work caught this: `hermes` exited 1 with `Unknown profile: dev` while every other adapter installed cleanly.

The script already creates a profile's `skills/` directory and its Claude-compatible `home/.claude/` underneath the profile, so refusing to create the profile root was inconsistent with its own behaviour. It now creates the directory and reports that it did.

`--all-profiles` is deliberately unchanged: an empty root there means there is genuinely nothing to enumerate, rather than one known target to create.

## Test plan

Verified both paths against a temporary `HERMES_PROFILES` root:

- [x] Default install into an **empty** root now creates `dev` and completes ("Done. Restart any running Hermes gateway…"), with the profile directory present afterwards
- [x] `--all-profiles` against an empty root still stops with "No profiles found under …"
- [x] `bash -n` syntax check passes
- [x] `tests/smoke.sh` passes


coco-research (approved):
Verified both paths against a temporary HERMES_PROFILES root: default install into an empty root creates dev and completes; --all-profiles still refuses an empty root.

PR #177, opened 2026-09-15
chore: ignore local tooling artefacts, stop restating a live context window

rijul-mck:
## Summary

Two small, unrelated housekeeping fixes.

### 1. Ignore local tooling artefacts

Two untracked directories, `.repowise/` and `coco-ads-output*` (and its dated siblings), repeatedly get swept into commits by a careless `git add -A`. Both contain absolute paths rooted in the operator's home directory, which trips the repository's push guard and blocks the push outright. This change adds `.repowise/` and `coco-ads-output*/` to the existing "Generated artifacts" section of `.gitignore`, alongside `node_modules/` and `__pycache__/`, matching the file's established style of one glob per line with a trailing slash for directories.

Verification performed in the worktree: created throwaway probe files under both directory patterns, confirmed `git status --porcelain` no longer lists them as untracked, and confirmed `git check-ignore -v` matches each one to the new `.gitignore` line:

```
.gitignore:31:.repowise/	.repowise
.gitignore:32:coco-ads-output*/	coco-ads-output
.gitignore:32:coco-ads-output*/	coco-ads-output-2026-09-15
```

Also confirmed the new rules do not over-match anything currently tracked: `git ls-files | git check-ignore --stdin` produced no output (exit code 1, meaning no tracked path matched).

### 2. Stop restating a live context window as documentation fact

`skills/local-llm/SKILL.md` documents a locally hosted language-model server. It hardcoded the server's context window as `131,072 (current default)` in a RAM-usage table, and repeated the same figure as settled fact in four other places in the file. The context window is a mutable runtime setting an operator changes by editing the server's own start script, so restating its value in documentation means the document goes stale the instant someone changes it.

Every restatement now instructs the reader to read the live value from the `context_window` field at the server's own health endpoint (`curl -s http://127.0.0.1:8090/health`), instead of asserting a fixed number. The RAM-usage table keeps its two illustrative rows (32,768 and 65,536 tokens, showing the linear KV-cache relationship) and adds prose explaining how to convert the live figure into the corresponding KV-cache estimate (divide by 16,384 for gigabytes per request, multiply by four for the `--max-batch 4` ceiling). One place that already stated a specific historical measurement (15.5 GB active / 18.6 GB peak, measured at a specific past context) keeps that number, but now frames it explicitly as "the context in effect when this was measured" rather than as the current default.

The server was unreachable during this session (`curl` to `http://127.0.0.1:8090/health` returned connection refused), so the `context_window` field name could not be independently re-verified against a live response. The field name used throughout this change is the one already established, unedited, elsewhere in the same file (in the pre-existing "Changing the context window" procedure and the `POST /admin/load` note); it was not invented for this change, but it is flagged here as unverified this session.

## Out of scope, deliberately

`assets/site/coco-advisory-board-389.png` renders stale asset-count figures and needs an image-regeneration pipeline to fix properly. Left untouched; noted here as known remaining work.

## Verification

All five required gates were run from this branch and passed:

```
$ python3 scripts/build-index.py
Wrote skills/INDEX.md
Wrote commands/INDEX.md
Wrote agents/INDEX.md
Wrote systems/INDEX.md
Wrote docs/by-domain/engineering.md
Wrote docs/by-domain/meta.md
Wrote docs/by-domain/design.md
Wrote docs/by-domain/pm.md
Wrote docs/by-domain/ops.md
Wrote docs/by-domain/foundational.md
Wrote docs/asset-counts.json

Done. Skills: 222 · Commands: 44 · Agents: 35
(exit 0, no diff against committed generated files)

$ bash tests/check-command-refs.sh
PASS: all command cross-references resolve.
(exit 0)

$ bash tests/check-evidence-gate.sh
PASS: /team evidence-gate language present.
(exit 0)

$ bash tests/check-asset-counts.sh
... all shipped public counts agree with docs/asset-counts.json
(exit 0)

$ bash tests/smoke.sh
=== Summary ===
  passed: 23
  failed: 0
(exit 0)
```

## Test plan

- [x] `git check-ignore -v .repowise coco-ads-output` matches the new rules
- [x] `git ls-files | git check-ignore --stdin` produces no output (no tracked file over-matched)
- [x] All five verification gates pass on this branch
- [ ] Reviewer confirms the `context_window` field name against a live `/health` response, since it could not be re-verified in this session


coco-research (approved):
Verified. The two .gitignore entries match only the intended paths, confirmed with git check-ignore -v, and git ls-files piped through git check-ignore --stdin returns nothing, so no tracked file is over-matched.

On the skill file, the arithmetic introduced is consistent with the two table rows that were retained: 32768 divided by 16384 is 2.0 and 65536 divided by 16384 is 4.0, and the removed row's own figures, 8.0 GB per slot and 32 GB across four slots, are reproduced exactly by the stated rule. The historical measurement is correctly re-attributed to the context in effect when it was taken rather than deleted or presented as current.

One limitation worth recording: the inference server was not running when this was written, so the context_window field name could not be confirmed against a live response. It is corroborated by the pre-existing and unedited troubleshooting section of the same file, which already refers to that field.

PR #178, opened 2026-09-15
fix(security): close the last contains-match uninstall glob and widen the gate

rijul-mck:
## Summary

This closes the one remaining instance of a defect that PR #117 mostly fixed, and widens the regression gate so the same class of bug cannot silently reappear under a different literal form.

### Defect: a contains-match glob in an uninstall instruction can delete the wrong links

PR #117 replaced a dangerous uninstall snippet of the form `find … -lname "*$(pwd)*"` everywhere it found it, because that pattern is a contains-match: a user uninstalling a clone at a path ending in `coco` would also delete symlinks pointing into a sibling directory such as `coco-research`, since the first path is a substring of the second. One instance survived the original sweep: `adapters/hermes/README.md` still told users to run `find ~/.hermes/profiles -type l -lname "*<path-to-coco-repo>*" -delete`, the same contains-match shape written with a placeholder instead of a shell substitution. This pull request rewrites it to resolve the clone path into a variable and anchor it as a prefix followed by a trailing slash, exactly matching the pattern already used by every other adapter's uninstall instructions (`claude-code`, `cursor`, `vscode`, the root `README.md`, and `docs/install.md`):

```bash
CLONE="$(pwd)"
find ~/.hermes/profiles -type l -lname "${CLONE}/*" -delete
```

**Adversarial proof.** I created a real `coco` clone directory and a sibling `coco-research` directory, each with its own file, and symlinked into both from a throwaway home directory (simulating what the Hermes adapter installs):

```
lrwxr-xr-x  bar -> .../coco-research/skills/bar
lrwxr-xr-x  foo -> .../coco/skills/foo
```

Running the corrected snippet from inside the `coco` clone root deleted only the link into `coco`:

```
$ ls -la home/.hermes/profiles/dev/skills
bar -> .../coco-research/skills/bar
```

`foo` (the link into the `coco` clone) is gone; `bar` (the link into `coco-research`) survives. I also confirmed the old contains-match pattern (`*<coco clone path>*`) matches both links, which is exactly why it was dangerous:

```
$ find home/.hermes/profiles -type l -lname "*$(cd coco && pwd)*"
.../home/.hermes/profiles/dev/skills/foo
.../home/.hermes/profiles/dev/skills/bar
```

### Gate: `tests/check-security-surface.sh` only caught one literal

The existing gate greps for the exact literal string PR #117 removed (`lname "*$(pwd)*"` in a fixed set of paths, plus two other specific literals), so it could not catch the Hermes survivor, which used different quoting and a placeholder instead of a shell substitution. The gate now also scans the whole tree for the general shape of the bug: an `-lname` argument that opens with a bare `*` before the path, rather than resolving the clone to a variable and anchoring it as a prefix. `CHANGELOG.md` is excluded because it is a historical record of the exact bug this suite guards against, and the test file itself is excluded because its own `grep` patterns quote the forbidden strings as literals in order to detect them elsewhere.

**Negative-control proof.** I temporarily reintroduced a contains-match variant in `adapters/hermes/README.md` (`-lname "*${CLONE}*"`, missing the trailing slash and opening with a bare asterisk) that none of the three pre-existing literal checks target. Running the widened gate:

```
=== uninstall: exact clone-path prefix, not contains-match ===
  PASS: bin/coco.js does not contains-match
  PASS: docs do not use *$(pwd)* contains-match
  PASS: vscode README does not use */coco/*
  PASS: find prefix deletes coco links, keeps coco-research
  FAIL: unanchored contains-match link glob found:
    ./adapters/hermes/README.md:49:find ~/.hermes/profiles -type l -lname "*${CLONE}*" -delete
```

The gate fails and names the exact file and line, confirming it catches the class rather than one literal. I then reverted the variant and reran the gate:

```
=== uninstall: exact clone-path prefix, not contains-match ===
  PASS: bin/coco.js does not contains-match
  PASS: docs do not use *$(pwd)* contains-match
  PASS: vscode README does not use */coco/*
  PASS: find prefix deletes coco links, keeps coco-research
  PASS: no unanchored contains-match link glob anywhere in the tree
```

### Two defects out of scope, confirmed absent from `main`

The originating brief also described two further defects to fix in `scripts/build-delivery-index.py` (a stale environment pop-list and a stale `ADVERTISED` count) and in a `pi-desktop` adapter detection probe. I verified both against the live tree rather than assuming: `scripts/build-delivery-index.py` does not exist anywhere on `main` (`git log --all` shows it was introduced only on the unmerged branch behind PR #169, `chore/arch-index-rebuild`, and inherited unchanged into PR #170). No `pi-desktop` adapter or `.pi-desktop` detection probe exists on `main` either, and it is confirmed present only on the unmerged PR #166, `feat/pi-desktop-adapter`, at `install.sh:39`. Neither is a defect in the code this pull request touches, and I did not invent new code to fix something that is not present on `main`, since doing so would duplicate and likely conflict with the versions already carried on those open pull requests.

## Test plan

- [x] `python3 scripts/build-index.py` — no drift in the generated index files.
- [x] `bash tests/check-security-surface.sh` — all checks pass, including the new class-wide check.
- [x] `bash tests/check-command-refs.sh` — pass.
- [x] `bash tests/check-evidence-gate.sh` — pass.
- [x] `bash tests/check-asset-counts.sh` — pass.
- [x] `bash tests/smoke.sh` — 23 passed, 0 failed.
- [x] Adversarial layout proof (coco vs. coco-research) for the corrected snippet, quoted above.
- [x] Negative-control proof (gate fails on a reintroduced variant, then reverted), quoted above.


coco-research (approved):
Verified. The adversarial case is real and the fix is correct: the previous pattern opened the -lname argument with a bare asterisk, making it a substring match, so a clone directory named coco also matched a sibling named coco-research. Anchoring on the clone path followed by a separator removes that.

The class-wide gate is the part that matters here, because the previous check only grepped for three exact literals and therefore could not catch a renamed variant. The two exclusions are justified in the code comment: CHANGELOG.md records the historical bug, and the test file itself necessarily quotes the forbidden strings in order to detect them.

Noting that this pull request's original brief named two further defects which do not exist on main. scripts/build-delivery-index.py and the pi-desktop probe live only on the unmerged pull requests 169, 170 and 166, so they are correctly handled there rather than invented here.

PR #179, opened 2026-09-15
feat(counts): derive persona, department and adapter counts, and gate them

rijul-mck:
## Summary

The repository published 389 personas across 9 departments in 29 locations across 9 files. The tree holds 495 personas across 13 departments: `find systems/superintelligence -path '*/personas/*.md' | wc -l` returns 495, `find systems/superintelligence -mindepth 2 -maxdepth 2 -type d -name personas | wc -l` returns 13. Every persona `.md` file under a `personas/` directory has real frontmatter (`slug`, `real_name`, `archetype`, ...); none is a README, index, or template file, so no exclusion rule was needed. The per-department sum (70+59+56+47+46+31+30+29+29+25+25+25+23) is exactly 495.

Four departments — Climate & Energy, Education & EdTech, Healthcare & Life Sciences, and Legal & IP — had real persona files on disk but were not wired into `registry.json` when this branch started. PR #138 (merged mid-rebase) wired all four in, converging "9 built teams" with "13 departments" — a distinction this branch had deliberately preserved up to that point (built-team-generated command math is a different count from persona-department headcount, and they had genuinely differed until #138 landed). That convergence raised the repo's skills total 222 → 226 (4 new `SKILL.md` files) and generated-commands total 242 → 342 (325 per-team at 25 verbs × 13 teams, + 17 meta), so this PR's own numbers were updated to match throughout, verified against `docs/asset-counts.json` after every rebase rather than assumed.

`ls adapters/` returns 16 directories today (14 at the time this branch opened; `grok` and `pi-desktop` landed via separate PRs mid-task), not the 9 that `.github/workflows/lint-frontmatter.yml`'s `KNOWN_ADAPTERS` allowed at the start.

## What this PR does

- `docs/asset-counts.json` now carries `personas.total`, `personas.by_department`, `departments.total`, `adapters.list` and `adapters.total`, all generated by `scripts/build-index.py` (`collect_personas()`, `collect_adapters()`) rather than hand-restated. Skills, commands and agents counting logic is unchanged; their totals moved only because #138 changed what's on disk.
- `tests/check-asset-counts.sh` now also gates:
  - **Persona/department claims** (`tests/check-persona-counts.py`) — README.md (hero line, badge, prose, department table with row-count and sum checks), package.json, agents/README.md, docs/install.md, docs/INDEX.md, systems/superintelligence/HOW-IT-WORKS.html, assets/og-image.svg.
[...]

PR #180, opened 2026-09-15
fix(ci): derive the adapter smoke-test loop and add a general drift gate

rijul-mck:
## Scope note

This pull request was narrowed mid-session after PR #179 was found to independently cover the specific `KNOWN_ADAPTERS` defect this work was originally briefed on. This pull request does not touch `.github/workflows/lint-frontmatter.yml` and does not fix `KNOWN_ADAPTERS`; that remains #179's change. What follows is the genuinely additional part: two other hardcoded adapter enumerations left unfixed by #179, and a general drift gate.

## Drift table (as found, before this change)

| File:Line | List | vs `ls adapters/` (16 real adapters) | Disposition |
|---|---|---|---|
| `.github/workflows/lint-frontmatter.yml:49-50` | `KNOWN_ADAPTERS = [...]` (10 entries) | Missing `vscode`, `hermes`, `aider`, `github-copilot-cli`, `vscode-continue` | Not touched here — fixed and gated by #179 |
| `tests/smoke.sh:30` | `for adapter in claude-code cursor grok vscode pi-desktop codex generic hermes; do` (8 entries) | Missing `aider`, `cline`, `roo-code`, `windsurf`, `zed`, `amazon-q`, `github-copilot-cli`, `vscode-continue` (8 of 16) | Fixed here: derived from `adapters/*/` |
| `.github/workflows/ci.yml:66` | same 8-entry loop as above | Same 8 missing | Fixed here: derived from `adapters/*/` |
| `bin/coco.js:106`, `bin/coco-bootstrap.sh:11`, `install.sh` header comments | `claude-code \| cursor \| grok \| vscode \| codex \| generic` (help text) | Partial by design, not a gate | Out of scope — prose usage examples, not a validation list |
| `scripts/build-delivery-index.py` | two further lists (per the brief) | N/A | Confirmed absent from `main`; exists only on unmerged #169/#170, not created here |

`ls adapters/` returns 16 directories. A `pi-desktop` adapter (PR #166) landed on `main` mid-session, adding itself by hand to the same two loops this pull request fixes — direct, fresh evidence that the hand-maintained loop keeps drifting and needs deriving, not just re-correcting once.

## What changed

- `tests/smoke.sh` and `.github/workflows/ci.yml`: the adapter dry-run smoke-test loop now iterates `adapters/*/` and derives the adapter name with `basename`, instead of restating a hand-picked subset. This is the same idiom `ci.yml`'s own "Bundle install gate" step already uses (`for manifest in adapters/*/manifest.json; do adapter=$(basename "$(dirname "$manifest")")`). The stale directory-existence guard and the "hermes ships via PR #118" comment are removed since they no longer apply once the loop iterates real directories.
- `tests/check-adapter-drift.sh` (new): a general gate, wired into `ci.yml` as its own step (`Validate no hardcoded adapter list drifts from adapters/`), matching how `check-command-refs.sh` and `check-evidence-gate.sh` are wired. It scans every file returned by `git ls-files` (restricted to `.sh`/`.py`/`.js`/`.yml`/`.yaml`, NUL-delimited) for two idioms — a bash `for <var> in <tokens>; do` loop, or a bracket/brace list literal — whose tokens are mostly (>=3 items, >=60% overlap) real adapter names, and fails when the tokens found disagree with `ls adapters/`.

**Coverage, stated plainly:** it will catch a brand-new hardcoded adapter list added later, in a file this script has never heard of, the same way it would have caught the two defects above. It does **not** catch a prose enumeration (a `claude-code | cursor | ...` help-text line, or a README compatibility table), a JSON file, or a list built from a variable/glob/concatenation rather than literal tokens — those aren't for-loops or bracket literals, and reliably parsing arbitrary prose without false positives isn't practical. It also deliberately excludes `.github/workflows/lint-frontmatter.yml`'s `KNOWN_ADAPTERS`: that list is owned and already gated by #179 under different semantics (subset-of, not equality, since its own comment allows it to list an adapter landing in its own PR ahead of disk), and scanning it here would just flag a known issue already being fixed elsewhere.

## A defect found in review, and fixed here in both places it existed

A reviewer ran this pull request's own detection logic against a real checkout and found it produced false positives: the checkout had three nested `git worktree`s under `.claude/worktrees/` (this project's own tooling creates them there) plus an untracked scratch file, and both `tests/check-adapter-drift.sh` (this PR) and the already-merged `tests/check-security-surface.sh` walked the filesystem (`pathlib.Path('.').rglob('*')`, and `grep -R --exclude-dir=.git`) rather than asking git what this repository actually tracks. `--exclude-dir=.git` does not reach into a nested worktree's checked-out files, since none of them are directories named `.git`; it only skips directories literally called that. So both gates scored content that is not part of this repository at all, and would fail in a checkout with nested worktrees on content that was already correct, while passing cleanly in CI, where a fresh checkout has none.

Both gates now enumerate candidates from `git ls-files` instead: NUL-delimited, so a path with a space or an unusual character can't be split wrong, and scoped to this worktree's own tracked index, which is what makes a nested worktree's checked-out copy invisible to the scan. `check-adapter-drift.sh` calls it via `subprocess` with the same extension/exclusion pathspec its header already documented; `check-security-surface.sh`'s unanchored-glob check now pipes it through `xargs -0 grep` instead of `grep -R`. The existing deliberate exclusions (`CHANGELOG.md`, each script excluding itself, and `check-adapter-drift.sh`'s `lint-frontmatter.yml` exclusion) are unchanged. `check-adapter-drift.sh`'s header comment is corrected to describe what the code does.

### Proof 1 — the false positive is gone (the decisive test)

Run from a real checkout containing three nested worktrees under `.claude/worktrees/` and an untracked `HANDOFF.md` (with the tracked content already carrying this pull request's fixes applied, since that checkout's own `main` branch does not have them yet):

```
### check-adapter-drift.sh ###
=== scanning tree for hardcoded adapter enumerations ===
  PASS: no hardcoded adapter enumeration disagrees with adapters/
EXIT=0

### check-security-surface.sh ###
  PASS: no unanchored contains-match link glob anywhere in the tree
=== Summary ===
  all security-surface checks passed
EXIT=0
```

Both pass, in the exact environment that previously produced the false positive.

### Proof 2 — real drift is still caught

Reintroduced the stale loop in tracked `tests/smoke.sh`:

```
FAIL: tests/smoke.sh:30 lists ['claude-code', 'codex', 'cursor', 'generic', 'grok', 'hermes', 'pi-desktop', 'vscode']
      missing from adapters/: ['aider', 'amazon-q', 'cline', 'github-copilot-cli', 'roo-code', 'vscode-continue', 'windsurf', 'zed']
```

Reverted; gate passed again (`PASS: no hardcoded adapter enumeration disagrees with adapters/`).

Separately, reintroduced an unanchored glob in tracked `docs/TROUBLESHOOTING.md`:

```
FAIL: unanchored contains-match link glob found:
    docs/TROUBLESHOOTING.md:14:find "$HOME" -type l -lname "*${CLONE}*" -delete   # TEST FIXTURE, remove before commit
```

Reverted; gate passed again (`all security-surface checks passed`).

### Proof 3 — untracked content is ignored

Created an untracked file at the repository root carrying both a stale adapter loop and an unanchored glob. Both gates still passed:

```
PASS: no hardcoded adapter enumeration disagrees with adapters/
PASS: no unanchored contains-match link glob anywhere in the tree
```

Deleted the file; both still pass.

## Verification (run from the branch, rebased onto current `main`)

```
$ python3 scripts/build-index.py
Done. Skills: 226 · Commands: 44 · Agents: 35
(no tracked file changed)

$ bash tests/check-command-refs.sh
PASS: all command cross-references resolve.

$ bash tests/check-evidence-gate.sh
PASS: /team evidence-gate language present.

$ bash tests/check-asset-counts.sh
=== Summary ===
  all shipped public counts agree with docs/asset-counts.json

$ bash tests/check-security-surface.sh
=== Summary ===
  all security-surface checks passed

$ bash tests/check-default-bundle-allowlist.sh
=== Summary ===
  all default-bundle-allowlist checks passed

$ bash tests/check-adapter-drift.sh
=== scanning tree for hardcoded adapter enumerations ===
  PASS: no hardcoded adapter enumeration disagrees with adapters/

$ bash tests/smoke.sh
=== Summary ===
  passed: 34
  failed: 0
```

All eight exit 0. `tests/smoke.sh`'s adapter dry-run loop exercises all 16 adapters, including the 8 that were silently skipped before (`aider`, `cline`, `roo-code`, `windsurf`, `zed`, `amazon-q`, `github-copilot-cli`, `vscode-continue`) — every one of their `install.sh --dry-run` calls passes.

`git diff origin/main...HEAD | grep '/Users/'` returns nothing.

## `runs-on` in `lint-frontmatter.yml`

Not modified. This pull request does not touch `.github/workflows/lint-frontmatter.yml` at all — confirmed with `git diff --stat` showing no entry for that file.

## Note on `.github/workflows/ci.yml`

PR #169 also modifies this file (a separate step, inserted after "Verify shipped asset counts", far from where this pull request inserts its own step). This pull request's change to `ci.yml` is kept to the minimum: one new step, and the adapter loop's body swapped for a derived one, both localized to a 13-line hunk, to reduce the conflict that landing both will cause.


coco-research (approved):
Verified. This fixes a defect in its own new gate and the same defect in one that had already merged, which is the right scope because they were one fault.

Both checks now build their candidate list from git ls-files with NUL delimiting, so a path containing a space cannot be split wrongly, and neither can see untracked content. That is the correct invariant: the property being enforced concerns the contents of this repository, not whatever happens to be checked out beside it. I confirmed the filesystem walk is gone from the drift gate entirely.

The decisive proof is the one run in a checkout that actually reproduces the failure, containing three nested worktrees under .claude/worktrees/ and an untracked file that quotes the forbidden pattern while documenting it. Both gates pass there now. Before this, the security-surface check reported six matches in that checkout and the drift logic produced eight hits of which six were nested-worktree copies, so both gates passed in continuous integration and failed only on a developer machine, which is the least visible failure mode and the one most likely to get a check switched off.

Real drift is still caught, proven separately for each gate by reintroducing the stale eight-adapter loop in a tracked file and an unanchored glob in tracked documentation, with each naming the exact file and line before being reverted. Untracked content carrying both defects is correctly ignored.

The underlying fix is also right in kind: the adapter dry-run loop is now derived from adapters/ rather than restated, and smoke coverage rose to 34 checks exercising all 16 adapters where 8 were previously skipped in silence. The evidence that a hand-maintained list keeps drifting is that the pi-desktop adapter added itself to that same loop by hand only hours ago.

PR #181, opened 2026-09-15
fix(ci): make frontmatter lint run on a hosted runner and cover the whole tree

rijul-mck:
## Summary

The `lint-frontmatter` workflow has never once executed successfully, and separately checked almost none of what it claims to lint. Both defects are fixed here; a third known defect (the `KNOWN_ADAPTERS` list) belongs to another in-flight pull request and is untouched.

### Why the check never ran

`runs-on` selected a self-hosted macOS runner for every push to `main` and every pull request from this repository (falling back to `ubuntu-latest` only for a fork's pull request). That self-hosted runner is online, but its `Set up Python` step fails on every run with:

```
##[error]mkdir: /Users/runner: Permission denied
##[error]The process '/opt/homebrew/bin/bash' failed with exit code 1
```

`actions/setup-python` writes its tool cache to a hosted-runner path that does not exist on that machine and cannot be created without root there. The job never reached a single validation step, in run `34983090132` or any other. Because the branch ruleset requires only `lint-and-test`, this failure never blocked anything, so the check has been decorative for its entire life.

### What changed about the runner, and why it is safe

`runs-on` is now the unconditional string `ubuntu-latest`. The job needs no local hardware: every step only reads files already checked out by `actions/checkout` and runs a pip-installed Python interpreter. The conditional this replaces existed because the repository is public with forks, and a fork's pull request could otherwise execute its own code on the maintainer's self-hosted machine. Moving unconditionally to a hosted runner removes that local-execution exposure entirely, for forks and for this repository's own branches and pull requests alike; it is strictly safer than the conditional it replaces, not a weakening of it. It is not the forbidden direction (flattening back to an unconditional self-hosted runner), which is why it is permitted here.

### Coverage, before and after

Before, the validation script globbed only `skills/*/SKILL.md` and `systems/*/skills/*/SKILL.md`, 208 files. It never checked `agents/`, `commands/`, the Superintelligence team catalog files at `systems/<bundle>/<team>/SKILL.md`, persona files, or adapter-specific skills, and its single-star patterns would have missed anything nested one level deeper still.

After, it checks 798 files across five kinds, with required fields drawn from what the real files in each kind consistently carry rather than an invented schema:

- Regular skill files (214: `skills/**/SKILL.md`, `systems/*/skills/**/SKILL.md`, `adapters/*/skills/**/SKILL.md`, each recursive): requires `name`, `description`.
- Superintelligence team catalog files (13: `systems/superintelligence/*/SKILL.md`): requires `last_updated`, the only field common to both schema variants in use (`team_id`/`team_name`/... vs `name`/`team`/...).
- Persona files (495: `systems/*/*/personas/*.md`): requires `slug`, `real_name`, `archetype`, `status`, all present in every file.
- Agent definitions (32: `agents/**/*.md`, excluding the all-caps documentation and preamble files `README.md`, `INDEX.md`, and the shared `PROMPT-DEFENSE.md`, none of which carry frontmatter): requires `name`, `description`.
- Slash commands (44: `commands/**/*.md`, excluding `INDEX.md`): requires `description` only, since a command's slash name comes from its file path, not its frontmatter (see `scripts/build-index.py`'s `collect_commands`), and only 5 of 44 command files carry a `name` field.

`workflows/**` remains a trigger path but is deliberately not given a frontmatter kind: none of its three files carry any frontmatter, so inventing a schema for them would fail every one.

### What the local run found

Running the widened check against the whole tree found exactly two genuine, mechanical problems:

- `agents/refactoring-specialist.md` was missing its frontmatter block entirely. Fixed by adding `name`/`description`, using the file's own existing descriptive paragraph. This also corrects its row in the auto-generated `agents/INDEX.md` (previously showing the include line `@agents/PROMPT-DEFENSE.md` as its description), produced by rerunning `scripts/build-index.py` as required by this repository's own CI gate that checks the generated index is not stale.
- `agents/vendor/claude-code-templates/architect-review.md` had an unquoted `description` whose embedded `user: '...'` text was parsed by YAML as a nested mapping key (`mapping values are not allowed here`). Fixed by quoting the value, matching every sibling agent file's convention; no content changed.

No other findings. Coverage was not narrowed to reach a green run.

### CI evidence

Confirmed the workflow YAML parses and `runs-on` is now a plain string (`ubuntu-latest`), reproduced every step's command locally in order (pip install of the pinned `pyyaml==6.0.2` in an isolated virtualenv, then the frontmatter validator, the adapter-manifest check, and the install-script syntax check), and ran `scripts/build-index.py` plus all six required test gates (`check-command-refs.sh`, `check-evidence-gate.sh`, `check-asset-counts.sh`, `check-security-surface.sh`, `check-default-bundle-allowlist.sh`, `smoke.sh`), all exit 0.

CI run `34985950822`, triggered by this pull request via the `pull_request` event, concluded `success`. The `lint` job's step list shows `Set up job`, `Run actions/checkout@v7`, `Set up Python`, `Install pyyaml`, `Validate frontmatter`, `Validate adapter manifests`, `Verify install scripts are executable`, and `Complete job` all green, on `pythonLocation: /opt/hostedtoolcache/Python/3.11.16/x64`, which is the hosted runner's Python, not the self-hosted machine's. The `Validate frontmatter` step's own output reads `Checked 798 files` followed by `All frontmatter valid.`, matching the local run exactly.

## Test plan

- [x] `.github/workflows/lint-frontmatter.yml` parses as YAML with `runs-on: ubuntu-latest`
- [x] Local reproduction of every workflow step, in order, all succeed
- [x] `python3 scripts/build-index.py` and all six `tests/check-*.sh` / `tests/smoke.sh` gates exit 0
- [x] `git diff origin/main...HEAD` contains no absolute `/Users/` paths
- [x] CI run `34985950822` on this pull request reached the `Validate frontmatter` step and concluded successfully


coco-research (approved):
Verified, including the one piece of evidence that actually settles this.

This check had never once executed successfully. On a same-repository event it selected a self-hosted runner where actions/setup-python writes its tool cache to the hosted runner's path and dies with mkdir: /Users/runner: Permission denied before reaching any validation step. So the job was decorative for its entire life, and because the ruleset requires only lint-and-test, nothing ever blocked on it.

The runner change is the right direction and worth being explicit about. The standing rule in this repository forbids flattening the conditional back to an unconditional self-hosted runner, because this repository is public with 11 forks and that would let a fork execute its own code on the maintainer's machine. Moving to an unconditional hosted runner is the opposite change: it removes local execution for forks and for this repository alike, so the exposure the conditional managed no longer exists. I confirmed runs-on is now a plain string and that the remaining steps need nothing from that Mac.

The coverage fix matters more than the runner fix. The validation previously globbed two shapes with single stars and checked 208 files, while the workflow triggered on agents and commands it never examined, and never looked at personas at all. It now checks 798 files across five kinds, with required fields established by inspecting real examples of each kind rather than by imposing a uniform schema, which is what keeps this from becoming a wall of failures that gets the check switched off again. Requiring only the field common to all 13 team catalog files is the correct conservative choice given two inconsistent schemas are in use there.

Two genuine defects were found and fixed as a consequence, an agent definition missing its frontmatter entirely and a vendored one whose unquoted description parsed as a nested mapping. That is a check earning its place on first contact.

Decisive evidence: run 34985950822 concluded success with the lint job reporting Checked 798 files and All frontmatter valid on the hosted runner. I also confirmed the adapter allow-list was genuinely left alone, with zero added or removed lines touching it.

PR #183, opened 2026-09-15
fix(adapters): install the 13 SI team front doors, and default every adapter to its bundles

coco-research:
## Two silent defects, found by auditing what an install actually delivers

You reported two skills missing from PI-Desktop. Both of those turned out to be installed — but chasing them surfaced two real gaps that were hiding behind the ones you noticed.

## 1. A third skill layout no adapter walked

`SKILL.md` lives in three places, and the repo's own `build-index.py` knows all three:

```
skills/<name>/SKILL.md
systems/<bundle>/skills/<name>/SKILL.md
systems/<bundle>/<team>/SKILL.md          <-- the Super Intelligence teams
```

Every adapter walked the first two. All **13 SI team front doors** were absent from every install:

```
before                              after
claude-code   175 of 188 skills     188
vscode        175 of 188            188
hermes        175 of 188            188
pi-desktop    178 of 191            191
```

A team front door is the skill that tells an agent what a team is for, which cells it has and how to convene it. The roster you can summon had no skill describing it. The folder is a short slug (`ai`, `gtm`) while the canonical id is the frontmatter name (`ai-super-intelligence`), which is what the generated index and the role roster already use.

## 2. Nine adapters never adopted the default-bundle contract

#170 made *every bundle by default* the rule. Nine adapters never got it, so a plain `bash adapters/<name>/install.sh` installed the core set alone:

| Adapter | Before | After |
|---|---:|---:|
| `grok` | **74** of 188 skills | 188 |
| `vscode-continue` | 74 of 148 | 148 |
| `windsurf` | 74 of 148 | 148 |
| `hermes` | 74 of 188 | 188 |

`grok` was the worst — 114 skills silently missing. It was also the adapter I had earlier used as the reference for handling team front doors: it implemented that layout correctly and still could not deliver it on a default run.

`hermes` needed a second fix. Its new front-door loop reused a `SKILL_COUNT` counter that `hermes` never initialises, and under `set -u` the increment aborted the install after one team. Caught by counting what landed rather than trusting the exit code.

## The gate that let this hide

`build-delivery-index.py` always passed `--systems`, so it measured the widest install and never the one a user runs.

It now measures **both** and prints a shortfall section. The moment it ran it named all three remaining adapters by itself:

```
**Adapters whose plain run falls short:**
| Adapter | Plain run | Widest | Missing | Exit |
| `grok` | 74 | 188 | 114 | 0 |
| `vscode-continue` | 74 | 148 | 74 | 0 |
| `windsurf` | 74 | 148 | 74 | 0 |
```

That section now reads *"Every adapter delivers its full skill set on a plain run"*, and CI holds it.

## Also

The measurement note explaining why skills stop below the advertised total named only the five Cursor-only skills. The larger share is the `reverse-skill` bundle — 33 skills of security tooling held back on purpose. The note now names both, so the gap reads as a decision rather than a defect.

## Verification

Full `ci.yml` mirror: **16/16 steps pass**, including the delivery-index gate and the bundle gate. Every adapter was installed into a throwaway `HOME` and counted, in both default and widest modes.

## Not changed

`aider` and `github-copilot-cli` are `AGENTS.md` producers and install no skill tree by design. `amazon-q`, `cline`, `roo-code` and `zed` install agents only, with no skills surface to default.


cococyracto (approved):
LGTM_SITE_ENG (Cyra / cococyracto)

- Closes real silent dogfood gap: 13 SI team front doors + default-bundle contract on adapters that missed #170.
- CI lint-and-test green.
- Delivery-index plain-vs-widest gate is the right honesty fix.

Note (non-blocking): adapters/INDEX plain-run table shows vscode-continue/windsurf short-by -27 (plain > widest) — follow-up measurement hygiene if real; does not block this install correctness fix.

Will not merge as coco-research / as approver-author. Owner merge.

PR #189, opened 2026-09-16
fix(team): use the placeholder the tooling actually expands

coco-research:
## The bug

`commands/team/ship.md` and `commands/team/_index.md` both end with:

```
ARGUMENTS: {{ARGUMENTS}}
```

Nothing expands that. Claude Code and PI-Desktop both substitute `$ARGUMENTS`, `$@` and `$1` — `{{ARGUMENTS}}` is not a placeholder in either, so the line renders literally as `ARGUMENTS: {{ARGUMENTS}}` no matter what was typed.

These are the two primary entry points, `/team` and `/team:ship`, which makes it the most visible instance of the pattern. The other **15 command files already use `$ARGUMENTS`**, and both of these files use `$ARGUMENTS` correctly in their own bodies — only the trailing line was wrong, which suggests it was copied as boilerplate and never exercised.

## Severity

Low. It breaks no pipeline: the router reads `$ARGUMENTS` earlier in the body, so the trailing line is redundant dead text. Worth fixing anyway, because it is dead text in the two commands a user is most likely to invoke, and because two conventions sitting side by side imply they are interchangeable when only one works.

## Verified

- No `{{ARGUMENTS}}` remains anywhere under `commands/`.
- `bash tests/check-command-refs.sh` passes.
- Full `ci.yml` mirror: **16/16 steps**. `lint-frontmatter.yml`: 3/3.


cococyracto (approved):
LGTM (cococyracto). Thin +2/−2: `{{ARGUMENTS}}` → `$ARGUMENTS` in commands/team/ship.md and _index.md only. Correct placeholder for tooling. Author coco-research ≠ approver. CI lint + lint-and-test green.

PR #191, opened 2026-09-17
fix(formula): pin Homebrew sha256 to the v1.5.0 tarball

coco-research:
## Summary
- Pin `Formula/coco.rb` sha256 to the live GitHub archive tarball for tag `v1.5.0`.
- Remove the leftover TODO that still pointed at the v1.2.0 hash.
- Version and URL were already `1.5.0` / `.../tags/v1.5.0.tar.gz` on `main` after Cyra's tag at `b806014`.

Computed from https://github.com/coco-research/coco/archive/refs/tags/v1.5.0.tar.gz (downloaded twice, same digest):

`c9a5a50017122f38dbd62fd096af61b8e87dd59928366341541a1cb9e0679cc9`

## Test plan
- [x] Downloaded the live archive URL tarball twice; sha256 matched both times.
- [x] Confirmed tarball prefix is `coco-1.5.0/`.
- [x] Confirmed `version "1.5.0"` and the v1.5.0 archive URL are unchanged.
- [ ] Homebrew audit/install is out of scope for this pin.

Do not merge from this PR author.


cococyracto (approved):
Cyra (cococyracto): Formula sha256 pin to v1.5.0 tarball only. CI green. Soft freeze.

PR #196, opened 2026-09-18
Team gate retrofit: scripts measure, hooks certify, prose judges

rijul-mck:
# Team gate retrofit: scripts measure, hooks certify, prose judges

## What this changes

The `/team:*` slash commands enforced their verification steps in prose that a model could follow or skip. This branch retrofits the pipeline so that every gate is measured by a Python script that writes a gate file and a receipt, every gate file is chained in an append-only receipt log the aggregators verify, and the command documents invoke those scripts and quote their output instead of describing the check. Judgement stays with the model; measurement moves to the scripts; certification moves to the hooks.

The work lives under `skills/team-gate/`:

- `scripts/gate_state.py` is the shared library: run directories under `~/.team/runs/<run-id>/`, an append-only `receipts.jsonl` with a hash chain, fourteen closed receipt kinds, two-way verification of every gate file and artifact a receipt names, and path containment for everything it resolves.
- `scripts/brownfield_map.py` writes the Stage 1 code map from ripwire, an offline tree-sitter based call-graph tool, so a run on an existing repository starts from a machine-made map before the model reads anything.
- `scripts/check_artifacts.py`, `run_gate.py`, `prove_red.py`, `verify_independent.py`, `claim_evidence.py`, `arch_gate.py` and the three renderers measure the stage handoffs, the CI commands and their parity, the red-green proof of every new test (including a stubbed-implementation proof for tests-only changes), the independent re-run in a clean worktree, the claims in the PR body against the evidence, the requirements matrix through the tests that back each requirement, and the architecture baseline, declared plan and conformance.
- `scripts/ship_gate.py` and `fix_gate.py` aggregate the gates into one verdict with a three-valued exit (0 pass, 1 block, 2 unrunnable), answer whether a stage may open, honour override receipts written in the human's own words, and block a fourth build round.
- `hooks/` holds four Node hooks with no dependencies that run in observe mode by default: the turn log records approvals and overrides from the human's literal words, the artifact guard asks the aggregator before a stage artifact is written and denies writes to evidence and gate files, the stage guard gates `gh pr create` and `git push` on the ship verdict, and the stop guard blocks a turn from ending mid-pipeline once. `hooks/install.py` registers them.
- `commands/team/{fix,test,develop,verify,reanalyse,ship}.md` now invoke the scripts by their installed path, state the sequence of actions between gates, and keep judgement prose word for word.

## How it was verified

Every script carries a `--self-test` that regenerates deterministic fixtures with pinned dates and runs the script's own CLI on temporary copies; `scripts/run_fixtures.sh` runs all fourteen suites and joins the existing CI step. The suite passed from a detached clean checkout at every milestone, most recently at 45e5dd2 with 14 of 14; one such run caught a regression a scoped review had missed, which is why the clean-checkout run stays the proof. Every task was built by one agent and verified by a separate agent that reproduced the claims live, and the verifier briefs and rulings are recorded in the handoff documents kept outside the repository.

Two pilot runs of the fix pipeline on a planted bug in a unittest repository, using the real ripwire binary, exercised the whole chain. Run 1 found five defects (the red classifier attributed a unittest assertion to the standard library frame, the claim checker read a flat evidence shape run_gate never wrote, the matrix could not map a requirement to any gate, a flag position, and a stale-discover rule the prose had not stated); each became a fixed and verified task. Run 2 passed every step with fix_gate reporting PASS, an intact chain, and a certified would-deny receipt for the deliberate attempt to hand-write the evidence file; an independent verifier confirmed it from the run directory alone. The ship pipeline was then executed end to end by two independent verifiers on a scratch repository, which exposed that the head rule blocked every pre-build gate after the build's commits; the rule is now scoped by stage (pre-build gates accept an ancestor head, built-tree gates require the exact HEAD) and the pipeline reaches PASS_WITH_OVERRIDE without rerunning any stage check.

## What is deliberately not in this branch

The hooks run in observe mode and record what they would have denied. Switching the denies on, one at a time, and observing a real pipeline under the harness needs a live Claude Code session started inside a pilot repository with an isolated configuration directory, which a subagent of the build session cannot exercise. Those steps are the next tasks and are documented in the hooks README. The command files invoke scripts at `~/.claude/skills/team-gate/scripts/`, matching the arch-index arrangement; that symlink is created after merge.

## Handoff for whoever merges and continues

`skills/team-gate/handoff/HANDOFF.md` is the latest brain handoff, timestamped 2026-09-18T13:02:30Z (UTC) and marked as superseding every earlier copy. Its section 0 gives the next agent, on any machine, the state at handoff, the ordered steps from clone through merge to the deferred tasks (the hook denies, the observed real run, the ripwire cross-check, the comparator, the small follow-ups), the invariants every change must keep, and the machine facts that do not travel. The register, the plan, the rulings, every brief and both pilot digests sit beside it in the same directory.

## Review notes

The commit history is one commit per verified task with full-prose messages. The five earliest commits below f3f028c are a first fixture generator's accidental commits, squashed into one honestly named commit before this PR was opened. `.team-ship/` and `.claude/` are ignored; nothing under either is committed.


coco-research:
CI is red on this branch, and it is a regression this branch introduced. Reporting with the exact fix rather than leaving it as "check failed", since it is a two-line change and it blocks the merge.

## What fails

```
FAILED STEP: Validate /team evidence-gate integrity
  MISSING in commands/team/test.md: protocol ref
  MISSING in commands/team/fix.md: fix protocol ref
```

Verified locally — `main` passes this gate, this branch does not:

```
origin/main                      PASS: /team evidence-gate language present.
origin/feat/team-gate-retrofit   MISSING in commands/team/test.md: protocol ref
                                 MISSING in commands/team/fix.md: fix protocol ref
```

## Why it happened

`tests/check-evidence-gate.sh` greps for two literal strings:

```bash
require "$T/test.md" "Test Evidence Protocol|team:evidence" "protocol ref"
require "$T/fix.md"  "team:evidence|Test Evidence Protocol" "fix protocol ref"
```

This branch replaced the prose references with the mechanism that supersedes them:

```diff
-Run the authoritative gate per the Test Evidence Protocol (`team:evidence.md`): ...
+python3 ~/.claude/skills/team-gate/scripts/run_gate.py evidence --repo-root .
+exit 0: continue and quote `.team-ship/EVIDENCE.md`; ...
```

That is the retrofit working as intended — prose replaced by a script with an exit code. But the literal strings left with the prose, so the old gate no longer finds them.

**The branch's own suite could not catch this.** `run_fixtures.sh` runs the team-gate self-tests; `check-evidence-gate.sh` is a different suite in the repo's CI. 14/14 on one and red on the other is exactly the shape of the defect this retrofitting effort exists to remove, so it is worth naming rather than quietly patching.

## Fix, option A — smallest, satisfies the gate (recommended)

Name the protocol alongside the new invocation, in both files:

```diff
--- a/commands/team/test.md
+++ b/commands/team/test.md
+The script implements the Test Evidence Protocol (`team:evidence.md`) so the gate is
+measured rather than narrated.

--- a/commands/team/fix.md
+++ b/commands/team/fix.md
+The script implements the Test Evidence Protocol (`team:evidence.md`) so the fix is
+graded on captured evidence.
```

Two lines. `UNVERIFIED` and `EVIDENCE.md` are still present in each file, so the gate's other two `require` checks already pass — only the protocol reference is missing.

## Fix, option B — evolve the gate instead

The gate is asserting a 2026-08-era convention: that a command file *names* the protocol in prose. Once `run_gate.py` is the enforcement, naming it is no longer the load-bearing thing — invoking it is. So the check could accept either form:

```bash
require_any "$T/test.md" "Test Evidence Protocol|team:evidence|run_gate.py evidence" "protocol ref"
```

I would not do this on this branch. It weakens a gate in the same pull request that first fails it, which is the pattern that makes a gate untrustworthy. Option A now, and if B is wanted, a separate pull request that argues its case on its own.

## Note on what would have caught it

The branch adds `run_fixtures.sh` to the standalone-suites step, which is good. But the step's other suites — `check-evidence-gate.sh`, `check-command-refs.sh`, `smoke.sh` — run *after* it in the same job, and a fixture suite that only covers its own directory cannot see a regression in `commands/team/`. If the retrofit produces a general drift gate, this is the case it should have caught.

coco-research:
Since the handoff bundle was committed, this branch has gained five commits that close
the deferred queue. Summary for review, newest last:

- `d111eec` restored the literal protocol reference that `tests/check-evidence-gate.sh`
  greps for in `commands/team/test.md` and `fix.md`. The retrofit had replaced that prose
  with `run_gate.py evidence` invocations, which is the intended direction, but the guard
  reads the string, so the branch was red on arrival.
- `d9383e4` made the arch fixture generator portable. It used BSD `sed -i ''`, which GNU
  sed reads as an empty script plus a file named after the substitution, so the build
  aborted on the Ubuntu runner.
- `487a8f1` gave the four drift fixtures the repository's own arch-index scripts. They had
  been resolving the tools from an installed copy under `~/.claude/skills/arch-index`,
  which exists on a developer machine and not on a fresh checkout, so four fixtures
  returned exit 2 there and the suite read 13/14.
- `367ab54` made the fixture runner print a failing suite's output. The two red runs above
  each reported only an exit code, which cost a local reproduction to diagnose.
- `7506c76` added per-guard hook arming (`hooks.stage`, `hooks.artifact`, `hooks.stop`),
  which is what the three deferred deny runs needed to isolate one guard at a time.
- `d66bbb1`, `4429861`, `e573d62`, `8414964` recorded the deny runs and closed the
  register follow-ups 5c, 5e, 5f, 9b (three of four ceilings), 12c, 12d and 23b.

Verification: `bash skills/team-gate/scripts/run_fixtures.sh` reports 14/14 both on the
Linux CI runner and on macOS at the same commit. Three defects were found by adversarial
verification after the builders' own suites reported success: `prove_red` skipped
`tearDown` exactly when a unittest test failed, the recorded `repo_root` rule was only
half aligned with `verify_chain` on malformed relative values, and the new unreadable
marker exception made eight scripts traceback with exit 1 instead of exiting 2.

Still deliberately open after this branch: register row 27 (the before/after comparator
for reanalyse, deferred by ruling R7) and the pytest fixture-aware runner and coverage
attribution ceilings under row 9b, both recorded in BUILD-REGISTER.md rather than marked
done.

coco-research (approved):
Board approve: CI green; ruleset needs 1 non-author write approve. Merge path: Rebase and merge (linear history; preserve per-task commits).

PR #198, opened 2026-09-20
chore(repo): Tier 1 Repo Standard

coco-research:
## What
Tier 1 only from `~/code/jev-use/docs/repo-playbook.html`: `AGENTS.md`, `.githooks/pre-push` + `core.hooksPath`, `.github/pull_request_template.md`.

## Why
Company law Repo Standard. Cyra week task.

## Verify
```
repo-check --repo <path> --tier 1
```

Soft-freeze. No Tier 2 stubs.

PR #199, opened 2026-09-20
docs(repo): complete the Repo Standard document set

coco-research:
## What this changes

Adds the five `docs/` documents that `AGENTS.md` already promised, fixes the PR template
filename, tracks `.metagpt/`, and ignores two generated artifacts. `repo-check` goes from
`NOT COMPLIANT, 3 required gaps` to `COMPLIANT`.

## Why

PR #198 landed `AGENTS.md` as a routing table naming six documents. Five of them did not
exist. The playbook's top rule is that a document that lies is worse than no document,
because the next agent acts on it, so the routing table was right and the documents were
missing. Tracked as task 1 in `docs/tasks.md`.

Three findings came out of running the audit rather than guessing:

1. **The PR template was there but invisible to the audit.** The tracked name was
   `.github/PULL_REQUEST_TEMPLATE.md`. macOS resolves the lowercase path on a
   case-insensitive filesystem, so a shell test says the file exists while `git ls-files`
   reports the uppercase name, and the standard matches lowercase.
2. **`.metagpt/` read as missing while sitting on disk.** The audit reads tracked files only,
   so `STATE.md` and `GATE.json` had to be committed. `GATE.json`'s `repo` field is now `.`
   rather than one machine's absolute path.
3. **A false positive in the new pre-push gate, found by wiring it.** The secret scan
   regex is `(api[_-]?key|secret|token).*=.*(sk-|ghp_|xox[baprs]-)`, and the committed
   ripwire map is a single 33 KB line containing both `est_tokens=` and the department name
   `risk-compliance`, whose substring `sk-` matches. The gate blocked the push over its own
   repository's content. I removed the trigger by not committing generated analysis output,
   which is the right call anyway, but the regex itself is still fragile and is now task 3 in
   `docs/tasks.md` with the exact reproduction.

## How it was verified

The audit, which is the standard's own definition of done:

```
$ repo-check
REQUIRED  6/6
EXPECTED  5/6
  GAP  no machine-specific paths  12 occurrence(s) across 10 file(s)
THE SIX DOCUMENTS  8/8
  pass  docs/prd.md          104 lines
  pass  docs/architecture.md 198 lines
  pass  docs/rules.md        195 lines
  pass  docs/design.md       159 lines
  pass  docs/tasks.md         62 lines
  pass  docs/memory.md       162 lines
  pass  .metagpt/STATE.md    starts with 'Next action:'
  pass  .metagpt/GATE.json   valid JSON
COMPLIANT  required tier clean. 1 improvement(s) available above.
```

The one reported gap is pre-existing and not mine: 12 machine paths across 10 files, in
skill docs and one CI comment. Left alone because fixing them is a different concern, and
because the check's own docstring calls it a judgement call. The 17 that were mine, in the
cross-check report merged with #196, are redacted to `<scratch>` and `~`.

Nothing broken, and the gate that now runs on push:

```
$ bash .githooks/pre-push
  ... secret scan
  ok  secrets
  ... npm test
  ok  test
GATE OK
```

```
$ for each of coco's fast gates
  pass  bash tests/check-command-refs.sh
  pass  bash tests/check-evidence-gate.sh
  pass  bash tests/check-roster-frontmatter.sh
  pass  bash tests/check-adapter-drift.sh
  pass  bash tests/check-asset-counts.sh
  pass  python3 systems/team/validate_roles.py
  pass  python3 systems/team/build_roster.py --check
  pass  no index drift after python3 scripts/build-index.py
SUMMARY: 7 pass, 0 fail
```

## Checklist

- [x] Gate is green (the hook ran on this push and printed `GATE OK`)
- [x] `docs/tasks.md` updated: it is new in this PR and carries this work, task 3, and the rest
- [x] `docs/memory.md` updated: it is new in this PR and records the decisions and dead ends
- [x] No secret, key, or personal path added; the 17 machine paths in the cross-check report were redacted
- [x] No documented invariant changed; `docs/rules.md` is new and describes invariants that already hold, each with its measurement

## Risk

**What could break.** The documents are new files and change no behaviour. The one
behavioural change is `.gitignore` gaining `.cartographer/`, `.metagpt/ripwire-*.txt` and
`.metagpt/ripwire-for.err`, so anyone who had those files staged by accident will now find
them ignored; that is the intent. `docs/design.md` records the site's current tokens, so a
site redesign will make it stale, and it says so in its own drift section.

**One thing the reviewer should know.** Wiring `core.hooksPath .githooks` is per clone and
cannot be committed. I ran it in this clone, which is shared with another agent working on
the site, so the gate now runs on their pushes too. It passes, but if it ever blocks
something legitimate, `git push --no-verify` is documented at the top of the hook, and the
fix belongs in task 3 rather than in a bypass.

**Rollback.** `git revert` of this commit restores the previous filenames and removes the
documents. Nothing here is load bearing for the build.


cococyracto (approved):
LGTM (cococyracto). Soft-freeze process unblock: branch updated onto main (985db8b / #201). Docs-only Repo Standard set. Author coco-research ≠ approver. Merge when CI green. Hold #183; #200 stays stacked on standard/repo-skeleton.

